PR #28, opened 2017-03-27
Implementing ays bridge

khaledkbadr:


templates/bridge/actions.py
```diff
@@ -1,6 +1,40 @@
...
+                       db=0,
+                       timeout=None)
+
+    # Create bridge
+    network_map = {
+        0: None,
+        1: 'static',
+        2: 'dnsmasq',
+    }
+    cl.bridge.create(service.model.data.name,
```
zaibon: There should be no `name` attribute in the schema the service itself already has a name. `service.name`

templates/bridge/actions.py
```diff
@@ -1,6 +1,40 @@
 
 def install(job):
-    raise NotImplementedError()
+    import g8core
+    service = job.service
+
+    # Get g8core client
+    node = service.parent
+    cl = g8core.Client(host=node.model.data.redisAddr,
```
zaibon: don't use g8core directly, use the client in jumpscale: 
`j.clients.g8core.get(host, port=6379, password='')`

templates/bridge/actions.py
```diff
@@ -1,6 +1,34 @@
...
+
+    # Create bridge
+    network_map = {
+        0: None,
+        1: 'static',
+        2: 'dnsmasq',
+    }
+    cl.bridge.create(service.name,
+                     hwaddr=service.model.data.hwaddr or None,
+                     network=network_map[service.model.data.networkMode.raw],
```
zaibon: you can do `str(service.model.data.networkMode)` so you don't need the `network_map` and the value comes from the schema directly

PR #34, opened 2017-03-28
configure the network of the node at boot

zaibon:


templates/network.g8/actions.py
```diff
@@ -0,0 +1,11 @@
+ def configure(job):
+    """
+    this method will be called from the node.g8os install action.
+    """
+
+    required_args = ['redisAddr', 'redisPort', 'redisPassword']
```
grimpy: woudlnt it be better to pass the node instance? which you can load here?

templates/bootstrap.g8os/actions.py
```diff
@@ -8,15 +8,17 @@ def bootstrap(job):
     job.logger.info("reverse ardb loopup on IP: {}".format(ip))
     mac = j.sal.nettools.getMacAddressForIp(ip)
 
+    service = job.service
     # create and install the node.g8os service
     node_actor = job.service.aysrepo.actorGet('node.g8os')
-    grid_config = job.service.aysrepo.servicesFind(actor='grid_config')[0]
+    networks = [n.name for n in service.producers.get('network', [])]
 
+    import ipdb; ipdb.set_trace()
```
grimpy: remove ipdb

templates/node.g8os/actions.py
```diff
@@ -25,12 +28,24 @@ def install(job):
...
-    # TODO
+    job.logger.info("configure networks")
+    loop = asyncio.get_event_loop()
+    futures = []
+    for network in service.producers.get('network', []):
+        job = network.getJob('configure', args={
+            'redisAddr': service.model.data.redisAddr,
+            'redisPort': service.model.data.redisPort,
+            'redisPassword': service.model.data.redisPassword,
+        })
```
grimpy: so pass node instance :D?

templates/network.g8/actions.py
zaibon: I can't pass object there cause the args are serialized

PR #35, opened 2017-03-29
Ays container

khaledkbadr:


templates/container/actions.py
```diff
@@ -1,13 +1,66 @@
 
 def install(job):
-    raise NotImplementedError()
+    service = job.service
+    node = service.parent
```
zaibon: duplicate as line 6

templates/container/actions.py
```diff
@@ -1,13 +1,66 @@
...
+                              port=node.model.data.redisPort,
+                              password=node.model.data.redisPassword)
+    # create ports config
+    ports = {}
+    if not service.model.data.ports:
+        ports = dict(map(int, port.split(':')) for port in service.model.data.ports)
+
+    # create bridges config
+    bridges = []
+    for bridge in service.model.data.bridges:
```
zaibon: For services consumed you should not take the names from the model and then do a servicesFind. Instead all the services comsumed are available in the `service.producers` dict. The key of this dict is the role of the service consumed and it return the list of producers for that role.

So here you should loop like this:
`for bridge in service.producers.get('bridge', []):`
bridge is already a Service object.

templates/container/actions.py
```diff
@@ -1,13 +1,66 @@
...
+                                       storage=service.model.data.storage or None)
+    service.model.data.id = container_id
+    service.model.data.status = 'running'
+
 
 def start(job):
-    raise NotImplementedError()
+    service = job.service
+    if str(service.model.data.status) == "halted":
+        service.executeAction('install', inprocess=True)
```
zaibon: this will never work, executeAction is a coroutine. you need to make it executed by an event loop.

There is a small tools in jumpscale that allow you to run asyncronous code:
```python
coro = service.executeAction('install')
j.tools.async.wrappers.sync(coro)
```

templates/container/actions.py
```diff
@@ -1,13 +1,66 @@
...
+                                  password=node.model.data.redisPassword)
+        cl.container.terminate(service.model.data.id)
+        service.model.data.status = 'halted'
+
 
 def monitor(job):
-    raise NotImplementedError()
-    # check if the container is still running, try to restart if it halted
+    service = job.service
+    service.executeAction('start', inprocess=True)
```
zaibon: same remark about asynchronous code.

templates/container/schema.capnp
```diff
@@ -4,16 +4,17 @@ struct Schema {
...
+    filesystems @5 :List(Text); # pointer to the filesystem to mount into the container [filesystem_name:/path/inside/container]
     zerotier @6 :Text; # pointer to the zerotier service to consume
     bridges @7 :List(Text); # pointers to the bridges to consumes
     hostNetworking @8 :Bool;
     # Make host networking available to the guest.
     # If true means that the container will be able participate in the networks available in the host operating system.
     ports @9:List(Text); # List of node to container post mappings. e.g: 8080:80
-
+    storage @10 :Text;
+    id @11: UInt32;
```
zaibon: good one ;-)

templates/container/schema.capnp
```diff
@@ -4,16 +4,17 @@ struct Schema {
     node @0 :Text; # pointer to the parent service
     status @1 :Status;
     hostname @2 :Text;
-    flist @3 :Text; Url to the root filesystem flist
+    flist @3 :Text; # Url to the root filesystem flist
     initProcesses @4 :List(Process);
-    filesystems: @5 :List(Text); # pointer to the filesystem to mount into the container
+    filesystems @5 :List(Text); # pointer to the filesystem to mount into the container [filesystem_name:/path/inside/container]
```
zaibon: Here we need to do differently, Cause for the consumption to work we need to have a list of instance of filesystem service that we consume. So we can put both the filesystem instance and the mountpoint in the container in the same line.

Maybe we could add a mounts property in the schema and use the trick you did here.
So the mounts would take a list of string with the format :
`filesystem_name:/path/inside/container`
and the filesystems property is just there for the consumption to work.

In that case it would be nice to implement a check in the input action and validate that all filesystem referenced in the mounts property are also present in the filesystems property.

templates/container/actions.py
```diff
@@ -1,13 +1,66 @@
...
+        if str(_bridge.model.data.networkMode) == 'dhcp':
+            bridges.append((_bridge.name, str(_bridge.model.data.networkMode)))
+        elif str(_bridge.model.data.networkMode) == 'static':
+            cidr = _bridge.model.data.setting.to_dict()['cidr']
+            bridges.append((_bridge.name, cidr))
+        else:
+            bridges.append((_bridge.name, ''))
+
+    mount_points = {}
+    for fs in service.model.data.filesystems:
```
zaibon: same remark regarding loop over producers and see down other remark about filesystems and mounts.

templates/container/actions.py
```diff
@@ -1,13 +1,66 @@
...
+    service = job.service
+    if str(service.model.data.status) == "halted":
+        service.executeAction('install', inprocess=True)
+
 
 def stop(job):
-    raise NotImplementedError()
+    service = job.service
+    if str(service.model.data.status) == "running":
+        node = service.parent
```
zaibon: duplicate as line 56

templates/container/actions.py
```diff
@@ -1,13 +1,66 @@
...
+            cidr = _bridge.model.data.setting.to_dict()['cidr']
+            bridges.append((_bridge.name, cidr))
+        else:
+            bridges.append((_bridge.name, ''))
+
+    # Create mount points mapping
+    mount_points = {}
+    for mount_point in service.model.data.mounts:
+        fs_name, container_mount_path = mount_point.split(':')
+        _fs = service.aysrepo.servicesFind(actor='filesystem', name=fs_name)[0]
```
zaibon: this is dangerous you don't test if servicesFind actually return anything.
instead use 
```python
try:
    _fs = service.aysrepo.serviceGet(actor='filesystem', name=fs_name)
except j.exceptions.NotFound:
    raise ...
```

templates/container/actions.py
```diff
@@ -1,13 +1,66 @@
...
+                                  port=node.model.data.redisPort,
+                                  password=node.model.data.redisPassword)
+        cl.container.terminate(service.model.data.id)
+        service.model.data.status = 'halted'
+
 
 def monitor(job):
-    raise NotImplementedError()
-    # check if the container is still running, try to restart if it halted
+    service = job.service
```
zaibon: Here you have to check if you can still reach the container with the client before trying to restart.
Then if you can't change the status and try to restart

PR #36, opened 2017-03-29
Ardb service

mie00:
Relies on the container service.

The commit was basically copied from here https://github.com/g8os/ays_template_g8os/blob/master/templates/ardb.g8os/actions.py

### DIFF
--- a/templates/ardb/actions.py
+++ b/templates/ardb/actions.py
@@ -4,7 +4,7 @@ def get_container_client_(service):
 def is_ardb_running_(client):
     try:
         for process in client.process.list():
-            if process['cmd']['arguments']['name'] == '/opt/bin/ardb-server':
+            if process['cmd']['arguments']['name'] == '/bin/ardb-server':
                 return process
         return False
     except Exception as err:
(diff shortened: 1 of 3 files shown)


templates/ardb/actions.py
```diff
@@ -1,16 +1,91 @@
...
+        if str(err).find("invalid container id"):
+            return False
+        raise
 
 def install(job):
-    raise NotImplementedError()
-    # write configuration file
+    import io
+    service = job.service
+    client = service.actions.get_container_client_(service=service)
```
grimpy: this underscore way of doing actions has been removed.
It's now possible to call all functions in the class normally

templates/ardb/actions.py
```diff
@@ -1,16 +1,91 @@
...
+    job.logger.info("update config")
+    content = content.replace('/mnt/data', service.model.data.homeDir)
+    content = content.replace('0.0.0.0:16379', '{host}:{port}'.format(host=service.model.data.host, port=service.model.data.port))
+
+    if service.model.data.master != '' and service.producers.get('master', None) is not None:
+        master = service.producers['master'][0] # it can only have one
+        content = content.replace('#slaveof 127.0.0.1:6379', 'slaveof {host}:{port}'.format(host=master.model.data.host, port=master.model.data.port))
+
+    # make sure home directory exists
+    client.filesystem.mkdir('service.model.data.homeDir')
```
grimpy: i assume this needs to be a variable not a string?

PR #37, opened 2017-03-30
Storagecluster

zaibon:


templates/storage_cluster/actions.py
```diff
@@ -1,19 +1,163 @@
...
+
+def create_ardb(storagecluster_service, fs_service, instance, master=None):
+    container_actor = storagecluster_service.aysrepo.actorGet('container')
+    ardb_actor = storagecluster_service.aysrepo.actorGet('ardb')
+
+    container_args = {
+        'node': fs_service.parent.parent.name,
+        'hostNetworking': True,
+        'filesystems': [fs_service.name],
+        'flist': 'https://hub.gig.tech/maxux/ardb-rocksdb.flist',
```
muhamadazmy: Once again @maxux proves indispensable. Shouldn't we have an organisation account for such base images?

templates/storage_cluster/actions.py
```diff
@@ -1,19 +1,163 @@
...
+        'storage': 'ardb://hub.gig.tech:16379',
+        'mounts': ["{}:{}".format(fs_service.name, "/mnt/data")],
+        'ports': ['16379:16379'],
```
muhamadazmy: that's totally ignored when `hostNetworking` is `True`

templates/storage_cluster/actions.py
maxux: This will be moved to `gig-official-apps` in couple minutes/hours.

templates/storage_cluster/actions.py
```diff
@@ -1,19 +1,163 @@
...
+    }
+    container_service = container_actor.serviceCreate(
+        instance=instance,
+        args=container_args
+    )
+
+    ardb_args = {
+        'homeDir': "/mnt/data",
+        'bind': '0.0.0.0:16379',  # FIXME: should be 40G network
```
muhamadazmy: If multiple containers are created on the same core0 host while they have `hostNetworking` they will get conflict ports of all start listening on `16379` (only the first will succeed and each other container will fail) so the port also must be configurable and make sure to not conflict with other running services

templates/storage_cluster/actions.py
```diff
@@ -1,19 +1,163 @@
...
+        'dataProfile': 'single',
+        'mountpoint': "/storage/{}".format(disk.name),
+        'devices': [disk.devicename],
+        'node': node_service.name,
+    }
+    sp_service = storagepool_actor.serviceCreate(instance=instance, args=sp_args)
+
+    fs_args = {
+        'storagePool': sp_service.name,
+        'readOnly': False,
```
grimpy: when my pr is done this will require a name argument aswell

templates/storage_cluster/actions.py
```diff
@@ -1,19 +1,163 @@
...
+        'quota': 0,
+        'mountpoint': "/storage/{}".format(disk.name),
```
grimpy: mountpoint should not be passed, altough its allowed, but we use conventien over configuration. after creasion the mountpoint will be set on the service model

templates/storage_cluster/actions.py
zaibon: indeed, forgot to deal with port number. thx

PR #46, opened 2017-04-04
VM and Volume Service implementation

zaibon:


templates/nbdserver/actions.py
```diff
@@ -0,0 +1,64 @@
...
+    # )
+    # FIXME: update when the change on the nbd server for grid API are done.
+    container.system(
+        '/nbdserver \
+        -protocol unix \
+        -address "/server.socket.{id}" \
+        -export {id}'
+        .format(id=service.model.key))
+
+    service.model.data.socketPath = 'nbd+unix:///{id}?socket=/server.socket.{id}'.format(id=service.model.key)
```
grimpy: this is not the socket path but the qemu/nbd protocol path, probably this path will be used from outside the container and will need to be prepended with the container rootfs location.
I would just store the path.

templates/nbdserver/actions.py
```diff
@@ -0,0 +1,64 @@
...
+    if process:
+        job.logger.info("killing process {}".format(process['cmd']['arguments']['name']))
+        client.process.kill(process['cmd']['id'])
+
+        job.logger.info("wait for nbdserver to stop")
+        for i in range(60):
+            time.sleep(1)
+            if is_running(client, service.model.key):
+                continue
+            return
```
grimpy: maybe at this point you can do a SIGKILL instead of SIGTERM?

templates/vm/actions.py
```diff
@@ -1,24 +1,182 @@
+def get_node_client(service):
+    node = service.parent
+    return j.clients.g8core.get(host=node.model.data.redisAddr,
+                                port=node.model.data.redisPort,
+                                password=node.model.data.redisPassword)
 
+def create_container(service, parent):
```
grimpy: when first reading this method I was confused. Maybe call it create_nbdserver_container ?

templates/vm/actions.py
```diff
@@ -1,24 +1,182 @@
...
+        service.consume(nbdserver)
 
 def install(job):
-    raise NotADirectoryError()
+    service = job.service
+
+    # get all path to the vdisks serve by the nbdservers
+    medias = []
+    for nbdserver in service.producers.get('nbdserver', []):
+        medias.append({'url': nbdserver.model.data.socketPath})
```
grimpy: so this is not correct cause you are pointing to a socketpath inside the rootfs of the container
so you need to get the container path something like /var/container/x/ and append it with the socketpath.

I beleive container info return this path

PR #53, opened 2017-04-05
api.raml adjustment for vm update

abdulrahmantkhalifa:
PUT node/<nodeid>/vm/<vmid>/volume/<volumesid


raml/api.raml
```diff
@@ -670,6 +670,18 @@ types:
...
                 application/json:
                   type: VM
+        put:
+          displayName: UpdateVm
+          description: updates the VM 
+          body:
+            application/json:
+              type: VMBase
+          responses:
+            200:
```
FastGeert: I don't see the point of returning this information

raml/types.raml
```diff
@@ -725,9 +719,9 @@ VMInfo:
   type: object
   description: Statistical information about a vm
   properties:
-    disk: VMDiskInfo[]
+    disks: VMDiskInfo[]
     net: VMNetInfo[]
-    cpu:
+    cpus:
```
FastGeert: cpus is not a collection, should be cpu

raml/types.raml
```diff
@@ -255,19 +255,14 @@ VMListItem:
...
-      type: string
-    status:
-      enum: [running, halted, paused, halting, migrating]
-      description: Status of the VM
+    name: string  
     memory:
       description: Amount of memory in MiB
       type: integer
-    cpu:
+    cpus:
```
FastGeert: Should be singular, we use plural only for collections

raml/types.raml
```diff
@@ -277,22 +272,21 @@ VM:
       description: List of virtual disks
       type: VDiskLink[]
 
+
+VM:
+  type: VMBase
+  properties:
+    id:
```
FastGeert: id should be in VMBase

raml/types.raml
```diff
@@ -725,9 +719,9 @@ VMInfo:
   type: object
   description: Statistical information about a vm
   properties:
-    disk: VMDiskInfo[]
+    disks: VMDiskInfo[]
     net: VMNetInfo[]
```
FastGeert: net --> nics

VMNetInfo --> VMNicInfo

raml/types.raml
```diff
@@ -300,6 +300,24 @@ VMCreate:
       description: Systemlevel cloudinit configuration
       type: object
 
+VMUpdate:
+  type: object
+  properties:
+    id: string
```
FastGeert: id not necessary here, will be in the url

abdulrahmantkhalifa: will it not change? i mean can't ovc need to change name ?

FastGeert: no changing id's is something we don't need

PR #54, opened 2017-04-05
Nodes api

zaibon:


.editorconfig
```diff
@@ -0,0 +1,13 @@
+root = true
```
zaibon: This file has nothing to do in the repo.

api/node/node_api_CreateBridge.go
```diff
@@ -0,0 +1,62 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	log "github.com/Sirupsen/logrus"
+	tools "github.com/g8os/grid/api/tools"
+	"github.com/gorilla/mux"
+)
+
+type bridgeBlueprint struct {
```
zaibon: if you don't reuse this type anywhere, don't define it in the global scope the package, but instead move it inside the function. I can even be a literal struct declaration.

api/node/node_api_CreateBridge.go
```diff
@@ -0,0 +1,62 @@
...
+	Node        string                      `json:"node"`
+}
+
+// CreateBridge is the handler for POST /node/{nodeid}/bridge
+// Creates a new bridge
+func (api NodeAPI) CreateBridge(w http.ResponseWriter, r *http.Request) {
+	var reqBody BridgeCreate
+
+	if err := json.NewDecoder(r.Body).Decode(&reqBody); err != nil {
+		w.WriteHeader(400)
```
zaibon: use the `tools.WriteError`

api/node/node_api_CreateBridge.go
```diff
@@ -0,0 +1,62 @@
...
+		return
+	}
+
+	// validate request
+	if err := reqBody.Validate(); err != nil {
+		w.WriteHeader(400)
```
zaibon: use the `tools.WriteError`

api/node/node_api_DeleteBridge.go
```diff
@@ -0,0 +1,25 @@
...
+
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/g8os/grid/api/tools"
+	"github.com/gorilla/mux"
+)
+
+// DeleteBridge is the handler for DELETE /node/{nodeid}/bridge/{bridgeid}
+// Remove bridge
+func (api NodeAPI) DeleteBridge(w http.ResponseWriter, r *http.Request) {
```
zaibon: Need to execute the delete action on the service before deleting it

api/node/node_api_DeleteStoragePool.go
```diff
@@ -0,0 +1,32 @@
...
+	"net/http"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/g8os/grid/api/tools"
+	"github.com/gorilla/mux"
+)
+
+// DeleteStoragePool is the handler for DELETE /node/{nodeid}/storagepool/{storagepoolname}
+// Delete the storage pool
+func (api NodeAPI) DeleteStoragePool(w http.ResponseWriter, r *http.Request) {
```
zaibon: Need to execute the delete action on the service before deleting it

api/node/node_api_DeleteFilesystem.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/g8os/grid/api/tools"
+	"github.com/gorilla/mux"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+// DeleteFilesystem is the handler for DELETE /node/{nodeid}/storagepool/{storagepoolname}/filesystem/{filesystemname}
+// Delete filesystem
+func (api NodeAPI) DeleteFilesystem(w http.ResponseWriter, r *http.Request) {
```
zaibon: Need to execute the delete action on the service before deleting it

api/node/node_api_GetBridge.go
```diff
@@ -0,0 +1,40 @@
...
+	if err := json.Unmarshal(srv.Data, &respBody); err != nil {
+		tools.WriteError(w, http.StatusInternalServerError, err)
+		return
+	}
+	respBody.Name = srv.Name
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	if err := json.NewEncoder(w).Encode(&respBody); err != nil {
+		log.Errorf("Error in encoding response: %+v", err)
+		tools.WriteError(w, http.StatusNotFound, err)
```
zaibon: you can't do that after the Headers has been written and you already sent data in the body.

api/node/node_api_GetStoragePools.go
```diff
@@ -0,0 +1,51 @@
...
+		go func(name string, i int) {
+			defer wg.Done()
+
+			schema, err := api.getStoragepoolDetail(name)
+			if err != nil {
+				log.Errorf("Error getting detail for storgepool %s : %+v\n", name, err)
+				return
+			}
+
+			respBody[i] = StoragePoolListItem{
```
zaibon: using append would allow to handle error better

api/node/node_api_ListBridges.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"github.com/g8os/grid/api/tools"
+)
+
+// ListBridges is the handler for GET /node/{nodeid}/bridge
+// List bridges
+func (api NodeAPI) ListBridges(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	nodeid := vars["nodeid"]
+	services, resp, err := api.AysAPI.Ays.ListServicesByRole("bridge", api.AysRepo, nil, map[string]interface{}{"parent": fmt.Sprintf("node.g8os!%s", nodeid)})
```
zaibon: can't we make that line smaller ?

api/node/node_api_ListBridges.go
```diff
@@ -0,0 +1,70 @@
...
+		tools.WriteError(w, http.StatusInternalServerError, err)
+		return
+	}
+	if resp.StatusCode != http.StatusOK {
+		w.WriteHeader(resp.StatusCode)
+		return
+	}
+
+	wg := sync.WaitGroup{}
+	var respBody = make([]Bridge, len(services), len(services))
```
zaibon: if you use append to fill the slice, you should not declare it with the size == len(service), only the capacity should be set to len(service).
So it should be `var respBody = make([]Bridge, 0, len(services))`

api/node/node_api_ListBridges.go
```diff
@@ -0,0 +1,70 @@
...
+	for _, service := range services {
+		go func(name, role string) {
+			defer wg.Done()
+
+			srv, resp, err := api.AysAPI.Ays.GetServiceByName(name, role, api.AysRepo, nil, nil)
+			if err != nil {
+				tools.WriteError(w, http.StatusInternalServerError, err)
+				return
+			}
+			fmt.Println(resp)
```
zaibon: remove debug code

api/node/node_api_ListBridges.go
```diff
@@ -0,0 +1,70 @@
...
+	var respBody = make([]Bridge, len(services), len(services))
+	wg.Add(len(services))
+
+	for _, service := range services {
+		go func(name, role string) {
+			defer wg.Done()
+
+			srv, resp, err := api.AysAPI.Ays.GetServiceByName(name, role, api.AysRepo, nil, nil)
+			if err != nil {
+				tools.WriteError(w, http.StatusInternalServerError, err)
```
zaibon: you can't write to the response in you loop.
Imagine all of the call to AYS fail. you will try to write the error multiple time.

The error need to be handle out of the loop.

Same apply for the code after

.editorconfig
khaledkbadr: It's for editors plugin `editorconfig` to unite format among editors. Nice to have but you can remove it if not needed

zaibon: your code should always be formated using gofmt. so I don't see the point

PR #76, opened 2017-04-13
Update build scripts

grimpy:


grimpy:
This does not make sense in the context of building, all builds start from a clean base withouth go

PR #118, opened 2017-04-24
g8os-network

mie00:
Two questions:
1. should this be called network.g8 or we keep or like this and remove network.g8.
2. why does this service has a node parent?


zaibon:
> should this be called network.g8 or we keep or like this and remove network.g8.

it should be called network.g8os

> why does this service has a node parent?

It should not, this service will be installed next to the bootstrap service during grid deployment.
and the bootstrap will consume these service to know which network it should make the node consume.

templates/network.g8os/actions.py
```diff
@@ -0,0 +1,64 @@
...
+        if any(addr['addr'].startswith(mgmtaddr + '/') for addr in nic['addrs'] if 'addr' in addr):
+            continue
+        if nic['speed'] == 0:
+            continue
+        interface = nic['name']
+        break
+    if interface is None:
+        raise j.exceptions.RuntimeError("No interface available")
+
+    ovs_container_id = int(node_client.container.create('https://hub.gig.tech/gig-official-apps/ovs.flist', host_network=True).get(0).data)
```
zaibon: since a node can consume multiple network services, you should check if an ovs container already exists and use it if it's the case. You can use container tags for that.

templates/network.g8os/actions.py
```diff
@@ -0,0 +1,68 @@
...
+
+def compine(ip1, ip2, mask):
+    """
+    >>> compine('192.168.1.11', '192.168.1.10', 24)
+    '192.168.1.11'
+    """
+    iip1, iip2 = [ip_to_dec(ip) for ip in (ip1, ip2)]
+    imask = int('1' * mask + '0' * (32 - mask), 2)
+    ires = (iip1 & (~ imask)) + (iip2 & imask)
+    return dec_to_ip(ires)
```
grimpy: combine is written with a b and why netaddr library is not used instead of doing this manually?

templates/network.g8os/actions.py
```diff
@@ -0,0 +1,68 @@
...
+
+    containers = node_client.container.find('ovs')
+    if containers:
+        ovs_container_id = int(list(containers))[0]
+    else:
+        ovs_container_id = int(node_client.container.create('https://hub.gig.tech/gig-official-apps/ovs.flist', host_network=True, tags=['ovs']).get(360).data)
+    container_client = node_client.container.client(ovs_container_id)
+    container_client.json('ovs.bridge-add', {"bridge": "backplane"})
+    container_client.json('ovs.port-add', {"bridge": "backplane", "port": interface, "vlan": 0})
+    node_client.bash('ip address add {addr}/{mask} dev backplane'.format(addr=storageaddr, mask=mask))
```
grimpy: please use system if bash is not required, not sure this flist will always have a shell included.

PR #149, opened 2017-05-01
Leverage g8os sal during vdisk/vm creation

grimpy:

### DIFF
--- a/templates/vm/actions.py
+++ b/templates/vm/actions.py
@@ -12,21 +12,9 @@ def input(job):
     return args
 
 
-def get_node_client(service):
-    node = service.parent
-    return j.clients.g8core.get(host=node.model.data.redisAddr,
-                                port=node.model.data.redisPort,
-                                password=node.model.data.redisPassword)
-
-
-def get_container_root(service, id):
-    client = get_node_client(service)
-    containers_info = client.container.list()
-    id = str(id)
-    if id in containers_info:
-        return containers_info[id]['container']['root']
-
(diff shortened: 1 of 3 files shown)


grimpy:
Please have a look at the idea, I still need to test this

templates/vm/actions.py
```diff
@@ -13,20 +13,8 @@ def input(job):
 
 
 def get_node_client(service):
```
zaibon: let's rename this to get node. so later in the code we avoid having `client.client.kvm...` and have instead `node.client.kvm.`

templates/vm/actions.py
```diff
@@ -95,24 +84,23 @@ def init(job):
...
     for nbdserver in service.producers.get('nbdserver', []):
         # build full path of the nbdserver unix socket on the host filesystem
-        container = nbdserver.parent
-        if container.model.data.id == 0:
+        container = Container.from_ays(nbdserver.parent)
+        if not container.is_running():
             # start container
-            j.tools.async.wrappers.sync(container.executeAction('start'))
-        container_root = get_container_root(service, container.model.data.id)
+            j.tools.async.wrappers.sync(nbdserver.parent.executeAction('start'))
```
zaibon: why not using the sal object and do `container.start()` ?

templates/vm/actions.py
grimpy: Sounds good!

templates/vm/actions.py
grimpy: to make sure the state in the ays model is set properly aswell

PR #161, opened 2017-05-04
container-info: add container nic info

mie00:


khaledkbadr:
Can you generate api.html please?

PR #162, opened 2017-05-04
Added monitoring service to check on node status

abdulrahmantkhalifa:


templates/node.g8os/actions.py
```diff
@@ -66,6 +66,17 @@ def install(job):
         loop.run_until_complete(asyncio.gather(*futures))
 
 
+def monitor(job):
+    from pprint import pprint; import ipdb; ipdb.set_trace()
+    service = job.service
+    addr = service.model.data.redisAddr
+    if j.sal.nettools.pingMachine(addr, pingtimeout=30):
```
grimpy: use ping method of g8os

templates/node.g8os/actions.py
```diff
@@ -66,6 +66,17 @@ def install(job):
         loop.run_until_complete(asyncio.gather(*futures))
 
 
+def monitor(job):
+    from pprint import pprint; import ipdb; ipdb.set_trace()
```
grimpy: remove pprint ipdb

PR #168, opened 2017-05-04
Pass config to nbdserver instead of gridapi

grimpy:
Adapt copyvdisk to pass metadata ardb


grimpy:
Dont merge untill ndbdserver changes are done

Issue #170, opened 2017-05-04
Add validation for enums


serboctor:
There is no validation done for any attribute that is expected to belong to a certain enum.
For example, in the bridge create, the networkMode is expected to be one of ["none", "static", "dnsmasq"]. If another value is supplied, the api doesn't validate and return a bad request which result in a crash in ays.

All payloads that contain enums should be validated and if a value other than the expected is supplied, the api should return a 400.

serboctor:
APIs affected by this fix:
- CreateContainer: nic type
- StoragePoolCreate: dataprofile and metadataprofile
- CreateVM: nics type
- UpdateVm: nics type
- VdiskCreate: vdisktype
- ClusterCreate:  drivetype

ahmedelsayed-93:
### SV
jumpscale : 68832ff8
grid : f100e6a

Issue #155, opened 2017-05-03
Need better validation of names and ids used for service names in ays


FastGeert:
I was struggling this morning fro more than 2 hours until i found out that i was using spaces in the api post field that underlying was going to be used as the service name in ays.

The grid api server just returns a 500, and in the body it returns 404, all very confusing.

In the log of the grid server I only found:
```
ERRO[0006] error executing blueprint for container Geert's echo service creation : 404 Not Found
192.168.193.33 - - [03/May/2017:07:46:23 +0000] "POST /nodes/0cc47a3b3d6a/containers HTTP/1.1" 500 26
```

serboctor:
I added a validation on the payloads of the apis to check that any attribute used as a service name matches the regex allowed by ays. Ays service names can only contain: alpha, numerics, -, _ and .
The APIs affected by this fix:
- BridgeCreate: name.
- CreateContainer: name.
- FilesystemCreate: name.
- StoragePoolCreate: name.
- VMCreate: id.
- VMMigrate: nodeid.
- ZeroTierJoin: nwid.
- CreateSnapshot.
- ClusterCreate: label.
- VdiskCreate: id.

ahmedelsayed-93:
### SV
jumpscale : 68832ff8
grid : f100e6a

Issue #152, opened 2017-05-02
bridges_apis: creating bridge with invalid parameters returns internal server error instead of 400


ahmedelsayed-93:
### API
```POST /nodes/{nodeid}/bridges```

this issue appears when i set ```name``` or ```networkMode``` or ``` hwaddr```  or ```settings > (cidr ,start, stop)```   with invalid value
* the invalid value for name parameter is special characters

serboctor:
Related to #155

serboctor:
Extra input from Ahmed:
- Invalid networkMode: any value except (none , static and dnsmasq)
- Invalid hwaddr: invalid mac address
- Invalid (cidr ,start, stop) wrong ip address format

ahmedelsayed-93:
### SV
jumpscale : 68832ff8
grid : f100e6a

PR #171, opened 2017-05-04
1.1.0 alpha validators

serboctor:
- Add validation for all enums #170
- Add validation for any attribute in a post payload that is used as an ays service name #155
- Add validation for some BridgeCreate attributes #152

### DIFF
--- a/api/validators/validators.go
+++ b/api/validators/validators.go
@@ -8,11 +8,11 @@ import (
 )
 
 func init() {
-	validator.SetValidationFunc("servicename", serviceName)
+	validator.SetValidationFunc("servicename", ServiceName)
 }
 
 // Validates that a string is a valid
-func serviceName(v interface{}, param string) error {
+func ServiceName(v interface{}, param string) error {
 	name := reflect.ValueOf(v)
 	if name.Kind() != reflect.String {
 		return errors.New("servicename only validates strings")
--- a/api/validators/validators.go
+++ b/api/validators/validators.go
@@ -1,0 +1,29 @@
+package validators
+
+import (
+	"errors"
+	"gopkg.in/validator.v2"
+	"reflect"
+	"regexp"
+)
+
+func init() {
+	validator.SetValidationFunc("servicename", serviceName)
+}
+
+// Validates that a string is a valid
+func serviceName(v interface{}, param string) error {
+	name := reflect.ValueOf(v)
+	if name.Kind() != reflect.String {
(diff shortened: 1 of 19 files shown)


api/validators/validators.go
```diff
@@ -0,0 +1,107 @@
...
+	if ip.Kind() != reflect.String {
+		return errors.New("ip only validates strings")
+	}
+
+	ipValue := ip.String()
+	if param == "empty" && ipValue == "" {
+		return nil
+	}
+
+	re, _ := regexp.Compile(`^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$`)
```
zaibon: Why not use [ParseIP](https://godoc.org/net#ParseIP) from std lib ?

api/validators/validators.go
```diff
@@ -0,0 +1,107 @@
...
+	if cidr.Kind() != reflect.String {
+		return errors.New("cidr only validates strings")
+	}
+
+	cidrValue := cidr.String()
+	if param == "empty" && cidrValue == "" {
+		return nil
+	}
+
+	re, _ := regexp.Compile(`^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(/([0-9]|[1-2][0-9]|3[0-2]))$`)
```
zaibon: Why not use [ParseCIRD](https://godoc.org/net#ParseCIDR) from std lib ?

api/validators/validators.go
```diff
@@ -0,0 +1,107 @@
...
+	if addr.Kind() != reflect.String {
+		return errors.New("macAddress only validates strings")
+	}
+
+	addrValue := addr.String()
+	if param == "empty" && addrValue == "" {
+		return nil
+	}
+
+	re, _ := regexp.Compile(`^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$`)
```
zaibon: Why not use [ParseMAC](https://godoc.org/net#ParseMAC) from std lib ?

api/validators/validators.go
serboctor: Google is no good. Where have you been all my life?

Issue #175, opened 2017-05-07
storageclusters_apis: delete storagecluster api doesn't work


ahmedelsayed-93:
### API
``` DELETE /storageclusters/{storageclusterid}```

### request & response
![image](https://cloud.githubusercontent.com/assets/24386820/25779790/a2596406-331e-11e7-860c-537f01de97ac.png)

**node disks status**
![image](https://cloud.githubusercontent.com/assets/24386820/25779801/df7a4242-331e-11e7-9319-abcedbfbe0f5.png)

### SV
jumpscale : 68832ff8
grid : f100e6a

ahmedelsayed-93:
![image](https://cloud.githubusercontent.com/assets/24386820/25848698/c848f37c-34bb-11e7-9f7f-a3b35ae52d5c.png)

![image](https://cloud.githubusercontent.com/assets/24386820/25848714/dc11fa34-34bb-11e7-89c8-44b5387fe2a6.png)

### SV
js 1a3b812
grid 67e7d28

abdulrahmantkhalifa:
https://github.com/g8os/resourcepool/pull/187

ahmedelsayed-93:
when deleting the storagecluster disk still mounted 
### SV
grid 48baa90
js d043e6e

ahmedelsayed-93:
### SV
js 441b344
resourcepool  da19c73

PR #177, opened 2017-05-07
completed the kill cluster api call to remove ardb, containers and storagepool

abdulrahmantkhalifa:
relates to issue #175

### DIFF
--- a/templates/storage_cluster/actions.py
+++ b/templates/storage_cluster/actions.py
@@ -66,9 +66,21 @@ def stop(job):
 
 def delete(job):
     service = job.service
-    cluster = get_cluster(service)
+    ardbs = service.producers.get('ardbs', [])
+    filesystems = service.producers.get('filesystems', [])
+
+    for ardb in ardbs:
+        container = ardb.parent
+        j.tools.async.wrappers.sync(container.executeAction('stop'))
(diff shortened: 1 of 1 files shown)


templates/storage_cluster/actions.py
```diff
@@ -66,9 +66,21 @@ def stop(job):
 
 def delete(job):
     service = job.service
-    cluster = get_cluster(service)
+    ardbs = service.producers.get('ardbs', [])
+    filesystems = service.producers.get('filesystems', [])
+
+    for ardb in ardbs:
+        container = ardb.producers['container'][0]
```
zaibon: Use parent relation here, not producers

templates/storage_cluster/actions.py
```diff
@@ -66,9 +66,21 @@ def stop(job):
...
+        j.tools.async.wrappers.sync(container.executeAction('stop'))
+        j.tools.async.wrappers.sync(container.executeAction('delete'))
+        j.tools.async.wrappers.sync(container.delete())
+
+    for fs in filesystems:
+        pool = fs.producers['storagepool'][0]
```
zaibon: Use parent relation here, not producers

Issue #180, opened 2017-05-08
Create container with nics type default fail 


Dinaamagdy:
## Steps : 
- Create container using API `Post /nodes/container' with nics type default and no value in nic Id as written in doc.

## Response: 
Bad request need to id .
![selection_317](https://cloud.githubusercontent.com/assets/24386802/25796722/fef05f1e-33da-11e7-9806-f70a9d6b3659.png)

## SV
 resourcepool : 5ab5ac39a6f84ec01ad98f1d9eb8235314a8c707

abdulrahmantkhalifa:
the grid no longer supports none type in the id , if u need to work pass either a zerrotierid vxlanid or vlantag

Dinaamagdy:
what is the Id value for default type ?

Dinaamagdy:
#SV 
resourcepool: 67e7d28a1207a57114a353093fd8261197c42d39

PR #182, opened 2017-05-08
bug fix concerning the default value of nic id

abdulrahmantkhalifa:
if type is default , id can be empty else
    id cannot be empty.

relates to issue #180

### DIFF
--- a/api/node/ContainerNIC.go
+++ b/api/node/ContainerNIC.go
@@ -1,8 +1,6 @@
 package node
 
 import (
-	"errors"
-
 	"github.com/g8os/resourcepool/api/validators"
 	"gopkg.in/validator.v2"
 )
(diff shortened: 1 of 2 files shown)


api/node/ContainerNIC.go
```diff
@@ -15,21 +17,35 @@ type ContainerNICConfig struct {
 type ContainerNIC struct {
 	Config ContainerNICConfig   `json:"config,omitempty" yaml:"config,omitempty"`
 	Hwaddr string               `json:"hwaddr,omitempty" yaml:"hwaddr,omitempty" validate:"macaddress=empty"`
-	Id     string               `json:"id" yaml:"id" validate:"nonzero"`
+	Id     string               `json:"id,omitempty" yaml:"id,omitempty"`
 	Type   EnumContainerNICType `json:"type" yaml:"type" validate:"nonzero"`
 }
 
+func (s ContainerNIC) validateID() error {
+	if s.Type != "default" {
```
serboctor: - Why are you using two if? why not just `if s.Type != "default && s.Id == """`?
- I think it is cleaner to use EnumContainerNICTypedefault instead of "default".
- Even thought this function is specific to ContainerNic but I feel like it should be part of the validators package. May be @zaibon can give us his input?

api/node/ContainerNIC.go
```diff
@@ -15,21 +17,35 @@ type ContainerNICConfig struct {
...
+		if s.Id == "" {
+			return errors.New("validation error on Id of NIC: cannot be empty")
```
serboctor: When I created our validators package, I tried to make the errors structure similar to the ones returned by validator.v2 package for the sake of consistency. Check [ValidateEnum](https://github.com/g8os/resourcepool/blob/1.1.0-alpha/api/validators/validators.go#L100) for reference.

PR #184, opened 2017-05-09
node-delete: add an ays action to delete a node

mie00:


raml/api.raml
```diff
@@ -28,6 +28,13 @@ types:
           application/json:
             type: Node[]
   /{nodeid}:
+    delete:
+      (AYSBased):
+      description: Delete a node
+      displayName: DeleteNode
+      responses:
+        204:
+          description: Jobs killed successfully
```
serboctor: Wrong description

templates/node.g8os/actions.py
```diff
@@ -93,3 +93,16 @@ def reboot(job):
         password=service.model.data.redisPassword or None,
     )
     node.client.raw('core.reboot', {})
+
+
+def uninstall(job):
+    service = job.service
+    bootstraps = service.aysrepo.servicesFind(actor='bootstrap.g8os')
+    if bootstraps:
+        import asyncio
```
zaibon: all this can be simplify by using `j.tools.async.wrappers.sync`

templates/node.g8os/actions.py
```diff
@@ -93,3 +93,16 @@ def reboot(job):
         password=service.model.data.redisPassword or None,
     )
     node.client.raw('core.reboot', {})
+
+
+def uninstall(job):
+    service = job.service
+    bootstraps = service.aysrepo.servicesFind(actor='bootstrap.g8os')
```
zaibon: need to deal with the case we don't find the service.

PR #200, opened 2017-05-10
1.1.0 alpha disk status

serboctor:


templates/storagepool/actions.py
```diff
@@ -107,3 +110,19 @@ def processChange(job):
             pool.ays.create(service.aysrepo)
         except ValueError:
             job.logger.error("pool {} doesn't exist, cant update devices", service.name)
+
+
+def monitor(job):
+    service = job.service
+    pservice = service.parent
```
zaibon: here you should only try to monitor when the install action has been done successfully.

PR #202, opened 2017-05-10
adjustments to delete in kill cluster + extra check on node to diskvalue

abdulrahmantkhalifa:
relates to issue https://github.com/g8os/resourcepool/issues/175


templates/storage_cluster/actions.py
```diff
@@ -147,6 +146,8 @@ def delete(job):
         j.tools.async.wrappers.sync(container.delete())
 
     for fs in filesystems:
+        if not fs.parent:
+            continue
```
grimpy: can you explain to me how this situation is possible?

abdulrahmantkhalifa: 2 filesystems can have the same pool , if its is deleted that pool will make the fs.parent of its child empty

PR #223, opened 2017-05-14
bug fix in migrate vm

abdulrahmantkhalifa:
relates to issue https://github.com/g8os/resourcepool/issues/215

### DIFF
--- a/templates/vm/actions.py
+++ b/templates/vm/actions.py
@@ -246,12 +246,12 @@ def migrate(job):
     service = job.service
 
     service.model.data.status = 'migrating'
-
-    args = job.model.args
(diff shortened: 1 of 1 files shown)


templates/vm/actions.py
```diff
@@ -380,13 +379,13 @@ def monitor(job):
 
 def update_data(job, args):
     service = job.service
-    service.model.data.memory = args['memory']
-    service.model.data.cpu = args['cpu']
+    service.model.data.node = args.get('node', service.model.data.node)
```
khaledkbadr: I think we pass the memory from the api whether it's updated or not

PR #235, opened 2017-05-16
Fix vm shutdown action.

serboctor:


templates/vm/actions.py
```diff
@@ -230,22 +230,23 @@ def resume(job):
 def shutdown(job):
     import time
     service = job.service
+    service.model.data.status = 'halting'
     job.logger.info("shutdown vm {}".format(service.name))
     node = get_node(service)
     kvm = get_domain(service)
     if kvm:
```
mie00: what happens if the domain is already halted, probably the status should be changed as well?

PR #240, opened 2017-05-17
Node disks refactor

serboctor:
- Change /node/nodeid/disks to list disks and their partitions and remove any data about mountpoints.
- Add /node/nodeid/mounts to list node mountpoints.


api/node/node_api_GetDiskInfo.go
```diff
@@ -27,30 +27,12 @@ func (api NodeAPI) GetDiskInfo(w http.ResponseWriter, r *http.Request) {
...
-		// disk information reflects info from any partition
-		for _, mount := range mounts {
-			if strings.HasPrefix(mount.Device, diskInfo.Device) {
-				diskInfo.Mountpoint = mount.Mountpoint
-				diskInfo.Opts = mount.Opts
-				diskInfo.Fstype = mount.Fstype
-				break
-			}
+		diskInfo := DiskInfo{
+			Device:     strings.Join([]string{"/dev", disk.Name}, "/"),
```
zaibon: `strings.Join` looks a bit overkill to me. maybe a simple `fmt.Sprintf` would do the trick ?

serboctor: Changed

PR #259, opened 2017-05-21
spwan one nbdserver per vm instead of one per vdisk

abdulrahmantkhalifa:
relates to issue https://github.com/g8os/resourcepool/issues/229

### DIFF
--- a/templates/nbdserver/actions.py
+++ b/templates/nbdserver/actions.py
@@ -6,12 +6,11 @@ def get_container(service):
     return Container.from_ays(service.parent)
 
 
-def is_running(container, key):
+def is_running(container):
     try:
         for process in container.client.job.list():
             arguments = process['cmd']['arguments']
-            if 'name' in arguments and arguments['name'] == '/bin/nbdserver' and \
-               key in arguments['args']:
+            if 'name' in arguments and arguments['name'] == '/bin/nbdserver':
                 return process
         return False
     except Exception as err:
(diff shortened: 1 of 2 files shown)


templates/nbdserver/actions.py
```diff
@@ -28,23 +28,33 @@ def install(job):
     from io import BytesIO
     from urllib.parse import urlparse
     service = job.service
+    vdisks = []
+    vm = service.aysrepo.serviceGet(role='vm', instance=service.name)
+    vdisks_names = j.data.serializer.json.loads(vm.model.dataJSON)['vdisks']
```
grimpy: this serliazes it to json to then unserialize it, better to use vm.model.data.to_dict()['vdisks']

templates/nbdserver/actions.py
```diff
@@ -28,23 +28,33 @@ def install(job):
...
-        config = container.node.client.config.get()
-        rootardb = urlparse(config['globals']['storage']).netloc
-    socketpath = '/server.socket.{id}'.format(id=service.name)
-    if not is_running(container, service.name):
+    config = {
+        'storageClusters': {},
+        'vdisks': {}
+    }
+    for vdiskservice in vdisks:
+        from pprint import pprint ; from IPython import embed ; import ipdb ; ipdb.set_trace()
```
grimpy: remove this line

templates/nbdserver/actions.py
```diff
@@ -53,21 +63,25 @@ def install(job):
...
-            -config {config}'
-            .format(id=service.name, socketpath=socketpath, config=configpath)
-        )
+        if  cluster.name not in config['storageClusters']:
+            config['storageClusters'][cluster.name] = clusterconfig
+        if rootclustername not in config['storageClusters']:
+            config['storageClusters'][rootclustername] = rootcluster
+        config['vdisks'][vdiskservice.name] = vdiskconfig
+
+    from pprint import pprint ; from IPython import embed ; import ipdb ; ipdb.set_trace()
```
grimpy: remove this line

templates/nbdserver/actions.py
```diff
@@ -28,23 +28,33 @@ def install(job):
...
+    for vdiskservice in vdisks:
+        from pprint import pprint ; from IPython import embed ; import ipdb ; ipdb.set_trace()
+        template = urlparse(vdiskservice.model.data.templateVdisk)
+        if template.scheme == 'ardb' and template.netloc:
+            rootardb = template.netloc
+        else:
+            conf = container.node.client.config.get()
+            rootardb = urlparse(conf['globals']['storage']).netloc
+        socketpath = '/server.socket.{id}'.format(id=service.name)
         configpath = "/{}.config".format(service.name)
```
grimpy: configpath and socketpath should probably be defined outside the forloop

templates/nbdserver/actions.py
```diff
@@ -28,23 +28,33 @@ def install(job):
...
         storageclusterservice = service.aysrepo.serviceGet(role='storage_cluster', instance=vdiskservice.model.data.storageCluster)
         cluster = StorageCluster.from_ays(storageclusterservice)
         clusterconfig = cluster.get_config()
```
grimpy: since this action is rather slow, you would better check here if vdisservie.model.data.storageCluster is already in config['storageClusters'] to avoid getting the config of the same cluster twice.

templates/nbdserver/actions.py
```diff
@@ -28,58 +28,77 @@ def install(job):
...
+    configpath = "/{}.config".format(service.name)
+    for vdiskservice in vdisks:
+        template = urlparse(vdiskservice.model.data.templateVdisk)
+        if template.scheme == 'ardb' and template.netloc:
+            rootardb = template.netloc
+        else:
+            conf = container.node.client.config.get()
+            rootardb = urlparse(conf['globals']['storage']).netloc
+
+        storageclusterservice = service.aysrepo.serviceGet(role='storage_cluster',
```
grimpy: move the next 2 lines under the if and change the if to
`if vdiskservice.model.data.storageCluster not in config['storageClusters']`

this should speed things up

templates/nbdserver/actions.py
```diff
@@ -28,58 +28,77 @@ def install(job):
     from io import BytesIO
     from urllib.parse import urlparse
     service = job.service
+    vdisks = []
+    vm = service.aysrepo.serviceGet(role='vm', instance=service.name)
+    vdisks_names = vm.model.data.to_dict()['vdisks']
```
zaibon: you can  replace this 3 line by directly accessing the `procuders` attribute
```python
vdisks = vm.producers.get('vdisks',[])
```

templates/nbdserver/actions.py
```diff
@@ -8,10 +8,10 @@ def get_container(service):
 
 def is_running(container, key):
     try:
-        for process in container.client.job.list():
-            arguments = process['cmd']['arguments']
-            if 'name' in arguments and arguments['name'] == '/bin/nbdserver' and \
-               key in arguments['args']:
+        for process in container.client.process.list():
```
grimpy: please change this back to job and call the variables job
you can now easily change if there is one job with arguments['name'] == '/bin/nbdserver' no need to check the args as there is only one nbdserver runnnning

templates/vm/actions.py
```diff
@@ -89,24 +81,25 @@ def init(job):
...
-        # build full path of the nbdserver unix socket on the host filesystem
-        container = Container.from_ays(nbdserver.parent)
-        if not container.is_running():
-            # start container
-            j.tools.async.wrappers.sync(nbdserver.parent.executeAction('start'))
-
-        # make sure the nbdserver is started
-        j.tools.async.wrappers.sync(nbdserver.executeAction('start'))
-        url = _nbd_url(container, nbdserver)
+    nbdserver = service.producers.get('nbdserver', None)[0]
```
zaibon: This will crash if there are not nbdserver consumed

abdulrahmantkhalifa: it should crash right , i just need to handle the error . becuse if there are no nbds at that point that means something went wrong right ?

PR #268, opened 2017-05-23
Add travis

grimpy:

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -2,7 +2,7 @@
 set -e
 pushd api
     echo "Build API"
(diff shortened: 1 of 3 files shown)


test.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+set -e
+pushd api
+    echo "Build API"
+    go get .
```
maxux: Should not be `go get ./...` ?

grimpy: Yip

PR #272, opened 2017-05-24
added the gateway create and list apis

abdulrahmantkhalifa:
also updated the html and ramls and the structs concerning gateway

### DIFF
--- a/api/node/node_api_CreateGW.go
+++ b/api/node/node_api_CreateGW.go
@@ -2,18 +2,59 @@ package node
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
+
+	"github.com/g8os/blockstor/log"
+	"github.com/g8os/resourcepool/api/tools"
+	"github.com/gorilla/mux"
 )
 
 // CreateGW is the handler for POST /nodes/{nodeid}/gws
 // Create a new gateway
 func (api NodeAPI) CreateGW(w http.ResponseWriter, r *http.Request) {
-	var reqBody GW
+	var reqBody GWCreate
+
+	vars := mux.Vars(r)
+	nodeID := vars["nodeid"]
 
 	// decode request
 	if err := json.NewDecoder(r.Body).Decode(&reqBody); err != nil {
 		w.WriteHeader(400)
 		return
 	}
+
+	// validate request
+	if err := reqBody.Validate(); err != nil {
+		tools.WriteError(w, http.StatusBadRequest, err)
+		return
+	}
+
+	gateway := struct {
+		Node         string        `json:"node" yaml:"node"`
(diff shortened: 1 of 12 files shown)


api/node/node_api_CreateGW.go
```diff
@@ -2,18 +2,59 @@ package node
...
 func (api NodeAPI) CreateGW(w http.ResponseWriter, r *http.Request) {
-	var reqBody GW
+	var reqBody GWCreate
+
+	vars := mux.Vars(r)
+	nodeID := vars["nodeid"]
 
 	// decode request
 	if err := json.NewDecoder(r.Body).Decode(&reqBody); err != nil {
 		w.WriteHeader(400)
```
serboctor: I always prefer using the status codes in http library e.g. http.StatusBadRequest instead of those generated numbers

api/node/node_api_ListGateways.go
```diff
@@ -2,12 +2,58 @@ package node
...
 func (api NodeAPI) ListGateways(w http.ResponseWriter, r *http.Request) {
-	var respBody []GW
+	vars := mux.Vars(r)
+	nodeID := vars["nodeid"]
+
+	query := map[string]interface{}{
+		"parent": fmt.Sprintf("node.g8os!%s", nodeID),
+	}
+	services, res, err := api.AysAPI.Ays.ListServicesByRole("gateway", api.AysRepo, nil, query)
+	if err != nil {
```
serboctor: You don't need this if and the next one. We have a HandleAysResponse in tools to handle repetitive code like this. Check listbridges as an example

api/node/node_api_ListGateways.go
```diff
@@ -2,12 +2,58 @@ package node
...
+	}
+	if res.StatusCode != http.StatusOK {
+		w.WriteHeader(res.StatusCode)
+		return
+	}
+
+	var respBody = make([]GW, len(services))
+	for i, serviceData := range services {
+		service, res, err := api.AysAPI.Ays.GetServiceByName(serviceData.Name, serviceData.Role, api.AysRepo, nil, nil)
+		if err != nil {
```
serboctor: Same comment as above.

api/node/node_api_ListGateways.go
```diff
@@ -2,12 +2,48 @@ package node
...
+		if !tools.HandleAYSResponse(err, res, w, "listing gatways") {
+			return
+		}
+		var data GW
+		if err := json.Unmarshal(service.Data, &data); err != nil {
+			tools.WriteError(w, http.StatusInternalServerError, err)
+			return
+		}
+
+		gateway := GW{
```
serboctor: Why are you initializing another GW variable instead of using data?

PR #276, opened 2017-05-24
added checks to validata ip versions and range

abdulrahmantkhalifa:


api/node/DHCP.go
```diff
@@ -10,6 +10,12 @@ type DHCP struct {
 }
 
 func (s DHCP) Validate() error {
-
+	if s.Hosts != nil {
```
grimpy: this is not needed in go, you can loop a nil list.

api/node/GWCreate.go
```diff
@@ -13,6 +13,26 @@ type GWCreate struct {
 }
 
 func (s GWCreate) Validate() error {
-
+	if s.Httpproxies != nil {
```
grimpy: same as for hosts

PR #285, opened 2017-05-25
added the update get and delete gw apis

abdulrahmantkhalifa:


api/node/GW.go
```diff
@@ -5,12 +5,28 @@ import (
...
+		if err := nic.Validate(); err != nil {
+			return err
+		}
+	}
+	for _, portforward := range s.Portforwards {
+		if err := portforward.Validate(); err != nil {
+			return err
+		}
+	}
+	return validator.Validate(s)
```
grimpy: double retunr?

PR #287, opened 2017-05-25
added the start and stop apis for the gateway

abdulrahmantkhalifa:


templates/gateway/actions.py
```diff
@@ -120,3 +120,31 @@ def processChange(job):
...
+    firewall = container.consumers.get('firewall')[0]
+
+    j.tools.async.wrappers.sync(container.executeAction('start'))
+    j.tools.async.wrappers.sync(http.executeAction('install'))
+    j.tools.async.wrappers.sync(dhcp.executeAction('install'))
+    j.tools.async.wrappers.sync(firewall.executeAction('install'))
+    j.tools.async.wrappers.sync(cloudinit.executeAction('install'))
+
+def stop(job):
+    from pprint import pprint ; from IPython import embed ; import ipdb ; ipdb.set_trace()
```
grimpy: ipdb?

templates/gateway/actions.py
```diff
@@ -120,3 +120,31 @@ def processChange(job):
...
+    service = job.service
+    container = service.producers.get('container')[0]
+    if container:
```
grimpy: why this if is here? Wouldnt you have crashed with a indexerror if it wasnt there?

PR #315, opened 2017-05-31
Rename pyclient to orchestrator

grimpy:


pyclient/zeroos/orchestrator/sal/abstracts.py
```diff
@@ -1,4 +1,4 @@
-from . import exceptions
+from JumpScale import j
```
muhamadazmy: why this is reintroduced ?

grimpy: it was a mistake ot create the other one, because we need to except NotFound from ays, so its not possible to create our own

PR #331, opened 2017-06-01
Update gateway definition

grimpy:

### DIFF
--- a/raml/types.raml
+++ b/raml/types.raml
@@ -1144,6 +1144,11 @@ GWNIC:
           default: id can be ommited
           bridge: name of the bridge to connect to
       type: string
+    zerotierbridge:
+      description: |
+        Zerotier NetworkId we want to use to bridge the private network to.
+        This makes it possible to bridge your private network with a private network of an other gateway
+      type: string
     config:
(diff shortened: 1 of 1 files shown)


raml/types.raml
```diff
@@ -1144,6 +1144,11 @@ GWNIC:
           default: id can be ommited
           bridge: name of the bridge to connect to
       type: string
+    zerotierbridge:
+      description: |
+        Zerotier NetworkId we want to use to bridge the private network to.
+        This makes it possible to bridge your private network with a private network of an other gateway
+      type: string
```
FastGeert: Please add the business rules to what the zerotier network should apply to

raml/types.raml
```diff
@@ -1152,7 +1157,9 @@ GWNIC:
            type: string
          gateway:
            type: string
-           description: Gateway for the the container
+           description: |
+             Gateway for the the container
+             When adding a gateway to a Nic it's considered a public interface
```
delandtj: errr... adding a gateway to a nic: what does that mean. Let's make sure we don't mix real-world terms to wrongly used...

Issue #338, opened 2017-06-02
Use YAML format for cloud-init configuration


FastGeert:
The current implementation expects the RESTful api user to embed json into the userdata & metadata fields, but that is very hard to document, because of lack of examples on the cloud init documentation site itself. They use YAML in all there examples, which is why we also need to use YAML.

ahmedelsayed-93:
### software version
orchestrator : bb0fb08

PR #344, opened 2017-06-04
Use yaml for cloud-init format #338

grimpy:

### DIFF
--- a/api/node/CloudInit.go
+++ b/api/node/CloudInit.go
@@ -1,7 +1,10 @@
 package node
 
 import (
+	"fmt"
+
 	"gopkg.in/validator.v2"
+	"gopkg.in/yaml.v2"
 )
 
 type CloudInit struct {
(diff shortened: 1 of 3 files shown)


api/node/CloudInit.go
```diff
@@ -10,6 +13,15 @@ type CloudInit struct {
...
-	return validator.Validate(s)
+	if err := validator.Validate(s); err != nil {
+		return err
+	}
+	data := make(map[interface{}]interface{})
+	if err := yaml.Unmarshal([]byte(s.MetaData), data); err != nil {
+		return fmt.Errorf("Invalid meta-data, could not parse YAML: %s %s", s.MetaData, err)
+	}
+	if err := yaml.Unmarshal([]byte(s.UserData), data); err != nil {
+		return fmt.Errorf("Invalid meta-data, could not parse YAML: %s %s", s.UserData, err)
```
serboctor: Wrong error message.

api/node/CloudInit.go
```diff
@@ -10,6 +13,15 @@ type CloudInit struct {
 }
 
 func (s CloudInit) Validate() error {
-
-	return validator.Validate(s)
+	if err := validator.Validate(s); err != nil {
```
serboctor: why not just do `return validator.Validate(s)` at the end of the file instead of this if statement and instead of the `return nil`.

PR #349, opened 2017-06-04
Configuration service

serboctor:
Implement configuration service to be able to specify certain versions of js, 0core, and flists.


templates/configuration/actions.py
```diff
@@ -0,0 +1,30 @@
+from JumpScale import j
```
grimpy: we are js9

Issue #335, opened 2017-06-01
Add shorthand endpoints for updating httpproxies in gw api


grimpy:
add

* Get /nodes/<nodeid>/gws/<gwname/httpproxes 
* POST /nodes/<nodeid>/gws/<gwname/httpproxes/<host>
* Delete /nodes/<nodeid>/gws/<gwname/httpproxes/<host>

khaledkbadr:
Done in https://github.com/zero-os/0-orchestrator/commit/217405a0b5627d7e72760563d21a97ce4bc9893e

Dinaamagdy:
Version: master @Revision: ffc0bd9e1e6da2e5c44fb126fcdeb5575eb8f169

PR #351, opened 2017-06-04
Add shorthand endpoints for updating httpproxies

khaledkbadr:
Implementation for #335


api/node/nodes_api_CreateHTTPProxies.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+// CreateHTTPProxies is the handler for POST /nodes/{nodeid}/gws/{gwname}/httpproxies
+// Create new HTTP proxies
+func (api NodeAPI) CreateHTTPProxies(w http.ResponseWriter, r *http.Request) {
+	var reqBody HTTPProxy
+
+	// decode request
+	if err := json.NewDecoder(r.Body).Decode(&reqBody); err != nil {
+		w.WriteHeader(400)
```
serboctor: I prefer `http.StatusBadRequest` than this generated number.

api/node/nodes_api_CreateHTTPProxies.go
```diff
@@ -0,0 +1,89 @@
...
+		errMessage := fmt.Errorf("Advanced options enabled: cannot add HTTp proxy for gateway")
+		log.Error(errMessage)
+		tools.WriteError(w, http.StatusForbidden, errMessage)
+		return
+	}
+
+	// Check if this proxy exists
+	var updatedProxies []HTTPProxy
+
+	if len(data.Httpproxies) > 0 {
```
serboctor: I don't get the point of the first if statement. And how is the updatedProxies correct? It doesnt contain the original proxies, does it?
What's wrong with doing the following? if the host already exists, the function will return inside the for loop. And if it doesn't exist, it will just be appended after the for lop.
```
for _, proxy := range data.Httpproxies {
			if proxy.Host == reqBody.Host {
				errMessage := fmt.Errorf("error proxy %+v already exists in gateway %+v", proxy.Host, gateway)
				log.Error(errMessage)
				tools.WriteError(w, http.StatusConflict, errMessage)
				return
			}
}
data.Httpproxies = append(data.Httpproxies, reqBody)
```

raml/api.raml
```diff
@@ -631,6 +631,48 @@ types:
...
+            get:
+              (AYSBased):
+              displayName: GetHTTPProxy
+              description: Get info of HTTP proxy
+              responses:
+                200:
+                  body:
+                    application/json:
+                      type: HTTPProxy
+            delete:
```
serboctor: Add the not found case.

api/node/nodes_api_CreateHTTPProxies.go
```diff
@@ -0,0 +1,84 @@
...
+		if proxy.Host == reqBody.Host {
+			errMessage := fmt.Errorf("error proxy %+v already exists in gateway %+v", proxy.Host, gateway)
+			log.Error(errMessage)
+			tools.WriteError(w, http.StatusConflict, errMessage)
+			return
+		}
+	}
+
+	var updatedProxies []HTTPProxy
+	data.Httpproxies = append(updatedProxies, reqBody)
```
serboctor: This is not correct. This will add the proxy in the reqbody to an empty slice which means the existing ones will be disregarded.
It should be `data.Httpproxies = append(data.Httpproxies, reqBody)`

PR #368, opened 2017-06-05
Fix install orchasterator script

khaledkbadr:


scripts/install-orchasterator.sh
```diff
@@ -1,5 +1,17 @@
 #!/bin/bash
 
+function valid () {
```
grimpy: use trap instead

scripts/install-orchasterator.sh
```diff
@@ -18,44 +30,58 @@ BRANCH=$1
 ZEROTIERNWID=$2
 ZEROTIERTOKEN=$3
 
-
+CODEDIR=/opt/code
```
grimpy: WTF? This should be doen from sourcing .jsenv is this variable is not set throw and error informing the user to set it or to source .jsenv manually

PR #391, opened 2017-06-06
Add configuration service to setup.md.

serboctor:


docs/setup/setup.md
```diff
@@ -183,6 +184,25 @@ pip3 install zerotier
...
+Configuring the jumpscale and 0-core version is possible with the [configuration](../versioning.md) service.
+
+For alpha-3, add the following blueprint in the `blueprint` directory of your AYS repository.
+ ```yaml
+ configuration__main:
+   configurations:
+   - key: '0-core-version'
+     value: 'v1.1.0-alpha-3'
+    - key: 'js-version'
+      value: 'master'
```
serboctor: I don't know if this should be `master` or should be something else. @grimpy  any input?

grimpy: there is no js9 verlease yet, so for now yes,

PR #411, opened 2017-06-07
Setup fix

FastGeert:


docs/setup/setup.md
```diff
@@ -86,10 +85,9 @@ After creating this blueprint, issue the following command to AYS to install it:
 ```
 (gig) root@js9:/root$ cd /optvar/cockpit_repos/orchestrator-server
 (gig) root@js9:/optvar/cockpit_repos/orchestrator-server$ ays blueprint network.bp
-(gig) root@js9:/optvar/cockpit_repos/orchestrator-server$ ays run create -y
 ```
 
-Then we need to update the bootstrap service so that is deploys the storage network when bootstrapping the nodes. So edit `/optvar/cockpit_repos/orchestrator-server/blueprints/network.bp` as follows:
+Then we need to update the bootstrap service so that is deploys the storage network when bootstrapping the nodes. So edit `/optvar/cockpit_repos/orchestrator-server/blueprints/bootstrap.bp` as follows:
```
yveskerwyn: Then we need to update the bootstrap service so that **is** deploys -> ... **it** deploys ...

Issue #374, opened 2017-06-06
vdisk status always halted 


Dinaamagdy:
## Issue
create vdisk , check its status , it always halted as shown : 

![selection_387](https://cloud.githubusercontent.com/assets/24386802/26819472/43cd4996-4a9f-11e7-9e26-ff6235dfaa9a.png)

## SV
f74e7be69b8594dba2a6b4d7c15933f25c3919b0

Dinaamagdy:
## Software Version
8cb8259f8d8b36008469d565d5ffe333e4c5be84

PR #417, opened 2017-06-07
added support for status change on vdisk

abdulrahmantkhalifa:
relates to issue https://github.com/zero-os/0-orchestrator/issues/374

Do not merge until issue of consumers is fixed


templates/nbdserver/actions.py
```diff
@@ -141,3 +148,17 @@ def stop(job):
                 continue
             return
         raise j.exceptions.RuntimeError("nbdserver didn't stopped")
+
+
+def monitor(job):
+    service = job.service
```
zaibon: check that the install state is `ok` before doing the monitoring

abdulrahmantkhalifa: done

Issue #348, opened 2017-06-04
Vdisk doesn't deleted after delete it's storagecluster


Dinaamagdy:
## Steps:
- Create storage cluster.
- Create Vdisk with it .
- delete this storage cluster 

## Expected result 
- vdisk be deleted 
## Issue
Vdisk exist  and be halted then If you try to delete it , you will have this error 

![image](https://cloud.githubusercontent.com/assets/24386802/26761111/d4418d74-4929-11e7-8fff-1bdb984651fd.png)

```
ays error 
 - schedule action delete on service:vdisk!ubuntu-test-vdisk5
init done
blueprint done
2017-06-04 11:21:06 - (network)[INFO][127.0.0.1:42954]: POST http://127.0.0.1:5000/ays/repository/resourcepool-server/blueprint/vdisk_ubuntu-test-vdisk5_delete_1496575265  200 71
2017-06-04 11:21:06 - (sanic)[ERROR]: Traceback (most recent call last):
  File "/opt/jumpscale8/lib/JumpScale/baselib/atyourservice81/lib/AtYourServiceRepo.py", line 365, in serviceGet
    return self.db.services.services[key]
KeyError: 'fde431de614f960494c21dcf51c6e0e4'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/sanic/app.py", line 471, in handle_request
    response = await response
  File "/opt/jumpscale8/lib/JumpScale/baselib/atyourservice81/server/ays_if.py", line 68, in post
    return await auth(request, ays_api.createRun(request, repository))
  File "/opt/jumpscale8/lib/JumpScale/baselib/atyourservice81/server/ays_if.py", line 16, in auth
    return await func
  File "/opt/jumpscale8/lib/JumpScale/baselib/atyourservice81/server/ays_api.py", line 257, in createRun
    to_execute = repo.findScheduledActions()
  File "/opt/jumpscale8/lib/JumpScale/baselib/atyourservice81/lib/AtYourServiceRepo.py", line 627, in findScheduledActions
    if is_producer_in_error(service, action):
  File "/opt/jumpscale8/lib/JumpScale/baselib/atyourservice81/lib/AtYourServiceRepo.py", line 606, in is_producer_in_error
    for producers in service.producers.values():
  File "/opt/jumpscale8/lib/JumpScale/baselib/atyourservice81/lib/Service.py", line 391, in producers
[...]

PR #423, opened 2017-06-07
Prevent storage cluster deletion if it has vdisks attached 

khaledkbadr:
#348

### DIFF
--- a/api/storagecluster/storageclusters_api_KillCluster.go
+++ b/api/storagecluster/storageclusters_api_KillCluster.go
@@ -27,7 +27,7 @@ func (api StorageclustersAPI) KillCluster(w http.ResponseWriter, r *http.Request
 
 	if len(services) > 0 {
 		err := fmt.Errorf("Can't kill nonempty storage clusters")
-		tools.WriteError(w, http.StatusInternalServerError, err)
+		tools.WriteError(w, http.StatusBadRequest, err)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


api/storagecluster/storageclusters_api_KillCluster.go
```diff
@@ -4,10 +4,9 @@ import (
 	"fmt"
 	"net/http"
 
-	"github.com/zero-os/0-orchestrator/api/tools"
 	"github.com/gorilla/mux"
-
-	log "github.com/Sirupsen/logrus"
+	"github.com/zero-os/0-disk/log"
```
serboctor: Don't we always use the logrus package?

api/storagecluster/storageclusters_api_KillCluster.go
```diff
@@ -16,6 +15,26 @@ func (api StorageclustersAPI) KillCluster(w http.ResponseWriter, r *http.Request
 	vars := mux.Vars(r)
 	storageCluster := vars["label"]
 
+	// Prevent deletion of nonempty clusters
+	query := map[string]interface{}{
+		"consume": fmt.Sprintf("storage_cluster!%s", storageCluster),
+	}
+	services, res, err := api.AysAPI.Ays.ListServicesByRole("vdisk", api.AysRepo, nil, query)
+	if err != nil {
```
serboctor: Don't we usually use `tools.HandleAYSResponse` to handle this if and the one after it?

PR #424, opened 2017-06-07
Bridge refactor validation

serboctor:


api/node/BridgeCreateSetting.go
```diff
@@ -24,6 +25,17 @@ type BridgeCreateSetting struct {
 }
 
 func (s BridgeCreateSetting) Validate() error {
+	if s.Cidr != "" && s.Start != "" {
```
grimpy: well start shouldnt be set if cidr isnt set

api/node/BridgeCreateSetting.go
```diff
@@ -24,6 +25,17 @@ type BridgeCreateSetting struct {
...
+		if err := validators.ValidateIpInRange(s.Cidr, s.Start); err != nil {
+			return err
+		}
+	}
+
+	if s.Cidr != "" && s.End != "" {
```
grimpy: end shouldnt be set if cidr isnt set

PR #442, opened 2017-06-09
Fix update of gateway

zaibon:


FastGeert:
Did you already test?

zaibon:
I close, can't managed to get any port forward created

Issue #445, opened 2017-06-11
Create methods should return run id in case possible


grimpy:
VM / container / gw / vdisk create (and maybe others) should return 202 Accepted instead of 201
The location header can remain pointing to the place the objects is going to be.

A new response will be added containing the runid.

A new method will be added on the orchestrator api to check the status of runs

GET /runs/< runid >

ahmedelsayed-93:
create storagecluster still returns 201 instead of 202 with run id

![image](https://user-images.githubusercontent.com/24386820/27281465-45d7b724-54ec-11e7-8737-678c52e03e54.png)

### Software version
orchestrator : master : db68214

FastGeert:
I do not agree with this change. @zaibon did you accept this. This is exactly what Kristof refused in my original design of the API.

FastGeert:
And what did we do with the status field?

zaibon:
We didn't change the behavior of the API, it was already like that, just that now we added a method to be able to track the status of the run.

I don't know anything about the status field,

FastGeert:
The fact that we use AYS as the deployment engine behind the API is irrelevant, and should not be exposed IMHO. What was the reason for adding this to the API?

grimpy:
This  feature has been rolledback. now all API's should be blocking untill the run is done.

PR #448, opened 2017-06-12
return runid in object create api calls

abdulrahmantkhalifa:
added the run get state and returned the run id in the
post requests of the objects
relates to issue #445


abdulrahmantkhalifa:
@zaibon , @grimpy  requested some changes , i did them

ahmedelsayed-93:
create storage cluster returns 201 not 202 with run id
image

Issue #449, opened 2017-06-12
Enable JWT for REST API


zaibon:
The orcestrator API has to be protected by JWT. 
The orchestrator is deployed for a specific IYO organization and only member of this organization can send request to the API.

Raml has to be updated with the `securedBy` node.
example : 
```raml
securedBy: [itsyouonline: { scopes: [ "user:memberof:organization_name" ] }]
```

Then the route interfaces has to be update to required the scope (https://github.com/zero-os/0-orchestrator/blob/master/api/node/node_if.go#L324)

you can generate the server in a temporary folder to see how go-raml creates it.

ahmedelsayed-93:
verification blocked on https://github.com/zero-os/0-orchestrator/issues/481

ahmedelsayed-93:
the duration of JWT token is 1 day, so if the owner of the organization remove user X form the organization,  user X will still be able to access the apis until the JWT token expire.
so we need to reduce the duration of the JWT token

grimpy:
@ahmedelsayed-93 altough a valid concern, this does not task wasnt properly implement this, please validate this ticket for the feature it implemnts, and create a new issue for your findings

PR #456, opened 2017-06-14
added itsyyouonline authentication and authorization

abdulrahmantkhalifa:
relates to issue #449


abdulrahmantkhalifa:
i added the docs for the http and the pyclient , but concerning the setup.md , it only talks about the docker install script , i did not update that script to use the jwt. As it is optional it does not affect the script running

docs/api.md
```diff
@@ -231,3 +231,51 @@ Payload:
...
+To generate the token the 0-orchestrator , the pyclient has provided an easy wrapper around the request that is used:
+```python 
+from zeroos.orchestrator.client import oauth2_client_itsyouonline
+
+cls = oauth2_client_itsyouonline.Oauth2ClientItsyouonline() # this class can take different urls to authenticate with but defaults to https://itsyou.online/v1/oauth/access_token?response_type=id_token
+
+response = cls.get_access_token(<client id>, <client secret>, scope=['user:memberOf:<organization name>'], audience=[]) # at the moment only the scope type organization:memberOf:<organization_name> is supported 
+
+print(respnose.token)
+```
```
FastGeert: typo in response

scripts/install-orchestrator.sh
```diff
@@ -22,17 +22,19 @@ export LANG=en_US.UTF-8
...
 
 if [ -z $1 ] || [ -z $2 ] || [ -s $3 ]; then
-  echo "Usage: installgrid.sh <BRANCH> <ZEROTIERNWID> <ZEROTIERTOKEN>"
+  echo "Usage: installgrid.sh <BRANCH> <ZEROTIERNWID> <ZEROTIERTOKEN> <ITSYOUONLINEORG>"
   echo
   echo "  BRANCH: 0-orchestrator development branch."
   echo "  ZEROTIERNWID: Zerotier network id."
   echo "  ZEROTIERTOKEN: Zerotier api token."
-  echo
+  echo "  ITSYOUONLINEORG:itsyou.online organizationfor use to authenticate."
```
FastGeert: Add a space after the colon

docs/setup/setup.md
```diff
@@ -30,8 +30,9 @@ So:
 (gig) root@js9:/root$ export BRANCH="1.1.0-alpha-3"
 (gig) root@js9:/root$ export ZEROTIERNWID="<Your ZeroTier network ID>"
 (gig) root@js9:/root$ export ZEROTIERTOKEN="<Your ZeroTier token>"
+(gig) root@js9:/root$ export ITSYOUONLINEORG="<Itsyou.online organization name>" #this is optional
```
FastGeert: Need to explain what this parameter is about.

FastGeert: This parameter is not and should not be optional. Please adapt the docs and also enforce it in the restapi server.

FastGeert: In the booting your nodes section of setup.md, you also need to add the itsyouonline parameter to the ipxe url for the nodes.

Also document there that it should be the same itsyou online org as here.

scripts/install-orchestrator.sh
FastGeert: Also document that nodes that get booted should use the same itsyou online organisation in its authentication parameter for accepting api requests to 0-core.

FastGeert: Something I can't find in the install script is where AYS gets configured with the itsyou online organisation to only accept REST calls authenticated with the correct JWT.

scripts/install-orchestrator.sh
```diff
@@ -122,7 +124,11 @@ fi
```
FastGeert: The installer should also configure AYS server to use the same itsyou online organisation for authenticating REST calls to AYS.

docs/setup/setup.md
abdulrahmantkhalifa: @grimpy  requested that the authentication be optional ,

api/tools/connection.go
```diff
@@ -107,17 +118,18 @@ func (c *connectionMiddleware) createPool(address, password string) *redis.Pool
...
+func (c *connectionMiddleware) getConnection(id NodeInfo, api NAPI) (client.Client, error) {
 	c.m.Lock()
 	defer c.m.Unlock()
 
-	if pool, ok := c.pools.Get(id); ok {
-		c.pools.Set(id, pool, cache.DefaultExpiration)
+	nodeId := id.id
+	token := id.token
+
+	if pool, ok := c.pools.Get(nodeId); ok {
```
grimpy: does not seem correct to me... how come you get it only by the nodeid? this means you might get a connection from someone else

abdulrahmantkhalifa: i create a pool per jwt

scripts/install-orchestrator.sh
abdulrahmantkhalifa: shouldnt the ays JWT configuration , be in the ays repo not here?
but if it needs to be here which file would that be

api/tools/connection.go
```diff
@@ -163,21 +175,28 @@ func ConnectionMiddleware(opt ...ConnectionOptions) func(h http.Handler) http.Ha
...
 	}
 
 	vars := mux.Vars(r)
-	id := vars["nodeid"]
+	token := strings.Split(r.Header.Get("Authorization"), " ")[1]
+	id := NodeInfo{token, vars["nodeid"]}
 
 	mw := p.(*connectionMiddleware)
-
+	mw.password = token
```
muhamadazmy: This doesn't look correct, setting a state on a global object and then pass credentials as method params ?!!
I can't trace the code where this is used. But doesn't look good to me.

api/run/runs_if.go
```diff
@@ -18,7 +20,7 @@ type RunsInterface interface { // GetRunState is the handler for GET /runs
 }
 
 // RunsInterfaceRoutes is routing for /runs root endpoint
-func RunsInterfaceRoutes(r *mux.Router, i RunsInterface) {
-	r.HandleFunc("/runs/{runid}", i.GetRunState).Methods("GET")
-	r.HandleFunc("/runs/{runid}/wait", i.WaitOnRun).Methods("GET")
+func RunsInterfaceRoutes(r *mux.Router, i RunsInterface, org string) {
+	r.Handle("/runs/{runid}", alice.New(tools.NewOauth2itsyouonlineMiddleware(org).Handler).Then(http.HandlerFunc(i.GetRunState))).Methods("GET")
```
muhamadazmy: U don't have to inject the middleware on each single handler. The idea of a middlewares is that they can be injected on higher entrypoints and then they become available for each handler. Check how the conneciton middleware is injected. 

This way it's really really hard to maintain

PR #463, opened 2017-06-14
review/update orchestrator setup documentation while  testing

yveskerwyn:


FastGeert:
@zaibon not really @yveskerwyn task to change these versions in the the documentation. I think this is something that needs to be part of the release process. Also, eg flists will not necessarily be updated when we push out a new release.

zaibon:
agree it's not his job
but if he touch these part of the file
I can't accept the PR

ok we keep reference to master on the master branch
and then I change all of them when creating the release branch

PR #484, opened 2017-06-18
Validate bridge used by container nics exists

grimpy:


api/node/GWCreate.go
```diff
@@ -27,10 +29,15 @@ func (s GWCreate) Validate() error {
 			return err
 		}
 	}
+	nicnames := make(map[string]struct{})
 	for _, nic := range s.Nics {
 		if err := nic.Validate(); err != nil {
 			return err
 		}
+		if _, exists := nicnames[nic.Name]; exists {
+			return fmt.Errorf("Duplicate nic names detected")
```
muhamadazmy: U don't have to change, but "error" messages in go should always start with `lowercase` letter as convention . like "duplicate nic names detected".

api/node/GWNIC.go
```diff
@@ -71,3 +73,16 @@ func (s GWNIC) Validate() error {
 
 	return validator.Validate(s)
 }
+
+func (s GWNIC) ValidateServices(aysClient tools.AYStool, repoName string) error {
```
muhamadazmy: Isn't a `GWNIC` already a `ContainerNic` ? i think duplicate code should have been avoided if GWNic extends `ContainerNic`

PR #503, opened 2017-06-20
installation: add tls to caddy

mie00:


scripts/install-orchestrator.sh
```diff
@@ -165,29 +185,33 @@ echo 'tmux new-session -d -s main -n 1 || true' >> ${orchinit}
...
+cat >> Caddyfile <<EOF
+$PUB {
     proxy / $PRIV:8080 {
         transparent
     }
+    $tls
 }
 
-http://ays.$DOMAIN:80 {
+$PUB {
```
grimpy: no need to proxy ays, also this would fail to start as its two times the same host

scripts/install-orchestrator.sh
```diff
@@ -165,29 +185,26 @@ echo 'tmux new-session -d -s main -n 1 || true' >> ${orchinit}
...
 echo 'tmux send-key -t orchestrator.0 "$cmd" ENTER' >> ${orchinit}
 
-if [ -n "$DOMAIN" ]; then
-    js9 'j.tools.prefab.local.apps.caddy.install()'
-    mkdir -p /opt/caddy
-    pushd /opt/caddy
-    cat >> Caddyfile <<EOF
-http://$DOMAIN:80 {
+js9 'j.tools.prefab.local.apps.caddy.install()'
+mkdir -p /opt/caddy
```
zaibon: I don't think this is the correct location for the caddy config, please check where jumpscale usually put these file and mimic behavior

scripts/install-orchestrator.sh
```diff
@@ -165,29 +185,26 @@ echo 'tmux new-session -d -s main -n 1 || true' >> ${orchinit}
...
+pushd /opt/caddy
+tls=
+if [ "$DEVELOPMENT" = true ]; then
+    tls='tls self_signed'
+fi
+cat >> Caddyfile <<EOF
```
zaibon: Need to check that the config is not already create, cause if we run the script twice, the Caddyfile will have duplicate content.

grimpy: can we just do a > instead of > or do we care about the config? Also is this script supposed to be rerunable?

zaibon:
Need to update setup.md document to explain the `DOMAIN` variable and `development` flag of the `install-orchestrator.sh` script

PR #505, opened 2017-06-20
Add ssh by default to ubuntu and clean ubuntu-minimal

maxux:


maxux:
Do not merge right now.

Issue #471, opened 2017-06-15
Revised storage setup


FastGeert:
For redudancy reasons we need to setup storage as follows:
![architecture](https://user-images.githubusercontent.com/15246294/27177335-4edec4a8-51c6-11e7-804a-e848176c972f.png)

So both the primary and backup storage cluster need to be setup with a metadata ardb server. Also the tlog server needs to be deployed on another machine if possible.

@robvanmieghem can you point the guys to the docs in 0-disk on how to configure the tlogserver process for the backup storage cluster connection.

Please use the storage-setup branch

robvanmieghem:
https://github.com/zero-os/0-Disk/blob/master/docs/tlog/config.md

Dinaamagdy:
can't validate due to :
https://github.com/zero-os/0-orchestrator/issues/662

Dinaamagdy:
## Software Version 
2b48d48e8d08470aaf9e8c2217ec60eff4122358

PR #522, opened 2017-06-21
Implement new architecture for tlogserver, vm, nbd with backup storage

khaledkbadr:
relates to #471


templates/tlogserver/actions.py
```diff
@@ -66,24 +82,30 @@ def install(job):
...
+                             log=logpath)
+            if backup:
+                cmd += '-with-slave-sync'
+            container.client.system(cmd)
             if not is_port_listening(container, port):
                 raise j.exceptions.RuntimeError('Failed to start tlogserver {}'.format(service.name))
             service.model.data.bind = '%s:%s' % (ip, port)
             container.node.client.nft.open_port(port)
+        else:
+            # send a siganl sigub(1) to reload the config in case it was changed.
```
grimpy: fix comment and use int(signal.SIGHUB)

templates/vm/actions.py
```diff
@@ -89,12 +91,23 @@ def _nbd_url(container, nbdserver, vdisk):
...
+    import random
     service = job.service
 
     # creates all nbd servers for each vdisk this vm uses
     job.logger.info("creates vdisks container for vm {}".format(service.name))
-    vdisk_container = create_zerodisk_container(job, service.parent)
-    _init_zerodisk_services(job, vdisk_container)
+    services = service.aysrepo.servicesFind(role="node")
+
+    node = random.choice(services)
```
grimpy: if there are miltiple nodes pop yourself and do random

templates/vm/actions.py
```diff
@@ -218,6 +231,21 @@ def stop(job):
     cleanupzerodisk(job)
 
 
+def destroy(job):
+    stop(job)
+    service = job.service
+    tlogservers = service.producers.get('tlogserver', None)
```
grimpy: default to empty list not None otherwise loop will crash

templates/vm/actions.py
khaledkbadr: That's what I do :p

Issue #515, opened 2017-06-21
Index page of the orchestrator API should not be protected by JWT


zaibon:
It should show a message that we reach the orchestrator API and a link to the api-console

zaibon:
fixed by #524

ahmedelsayed-93:
### software version
master : 92d0ba0

PR #524, opened 2017-06-21
Address issue #515

grimpy:


api/apidocs/index.html
```diff
@@ -1,17 +0,0 @@
-<!doctype html>
```
zaibon: This need to be a real home page. with a minimum css too so it's not to ugly

PR #577, opened 2017-07-03
Stats collector

serboctor:


templates/stats_collector/actions.py
```diff
@@ -0,0 +1,103 @@
+def get_init_processes(service):
+    from zeroos.orchestrator.configuration import get_jwt_token
+
+    token = get_jwt_token(service.aysrepo)
+    cmd_args = []
+
+    if token:
+        cmd_args.extend(['-- token', token])
```
grimpy: should be `--token` withouth the option

PR #611, opened 2017-07-04
Initial grafana service commit

mie00:


templates/grafana/actions.py
```diff
@@ -0,0 +1,122 @@
...
+    service.saveAll()
+
+
+def uninstall(job):
+    service = job.service
+    containers = service.producers.get('container')
+    if containers:
+        j.tools.async.wrappers.sync(containers[0].executeAction('stop', context=job.context))
+        j.tools.async.wrappers.sync(containers[0].delete())
+    service.delete()
```
zaibon: this call is also a coroutine, it needs to be wrapped in a `j.tools.async.wrappers.sync`

PR #614, opened 2017-07-04
reinstall node on reboot

mie00:


templates/node.zero-os/actions.py
```diff
@@ -64,6 +75,11 @@ def monitor(job):
     from zeroos.orchestrator.configuration import get_jwt_token
     import redis
     service = job.service
+
+    installed = isInstalled(job)
+    if not installed:
+        install(job)
```
grimpy: its best to call this with the syncwrapper to make sure the audit trail is complete in showing the fact that install was called from the monitor

templates/node.zero-os/actions.py
```diff
@@ -42,6 +42,17 @@ def getAddresses(job):
     return networkmap
 
 
+def isInstalled(job):
+    from zeroos.orchestrator.sal.Node import Node
+    from zeroos.orchestrator.configuration import get_jwt_token
+
+    service = job.service
+    node = Node.from_ays(service, get_jwt_token(job.service.aysrepo))
+    job.logger.info("mount storage pool for fuse cache")
```
grimpy: remove this log message as its not true

Issue #634, opened 2017-07-05
Create script that generate storagecluster dashboard


grimpy:
We need to have a script (later this needs to be automated) that generates a dashboard for a storagecluster.

This dashboard needs to contain a view on the disk usage and IOPS of all disks involved in the storagecluster.
Invidual and agregate view should be possible.

**Scenarios:**
* `statsdb` service is already deployed. Then when creating storage cluster it should generate its dashboard.
* `storage_cluster` already deployed. Then when deploying statsdb service it should also generate dashboard of storage clusters found.
* statsdb is not deployed at all. Then when deploying storage cluster dashboard shouldn't be generated at all

Dinaamagdy:
-After run statsdb service  ,can't create storagecluster with this error :
```
[Wed12 10:09] - Job.py              :57  :j.core.jobcontroller.job.7b134b012036611b21a7cdd0db1d55dc - ERROR    - Traceback (most recent call last):
  File "/usr/lib/python3.5/concurrent/futures/thread.py", line 55, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/tmp/actions/storage_cluster/af80d71e4b39316d7caa3449407d309a.py", line 169, in init
    cluster = get_cluster(job)
  File "/tmp/actions/storage_cluster/af80d71e4b39316d7caa3449407d309a.py", line 71, in get_cluster
    return StorageCluster.from_ays(job.service, job.context['token'])
  File "/usr/local/lib/python3.5/dist-packages/zeroos/orchestrator/sal/StorageCluster.py", line 36, in from_ays
    storages_server = StorageServer.from_ays(storageEngine_service, password)
  File "/usr/local/lib/python3.5/dist-packages/zeroos/orchestrator/sal/StorageCluster.py", line 157, in from_ays
    storageEngine = StorageEngine.from_ays(storageEngine_services, password)
  File "/usr/local/lib/python3.5/dist-packages/zeroos/orchestrator/sal/StorageEngine.py", line 30, in from_ays
    container = Container.from_ays(service.parent, password)
  File "/usr/local/lib/python3.5/dist-packages/zeroos/orchestrator/sal/Container.py", line 96, in from_ays
    sp = node.storagepools.get(fs_service.parent.name)
  File "/usr/local/lib/python3.5/dist-packages/zeroos/orchestrator/sal/StoragePool.py", line 59, in get
    raise ValueError("Could not find StoragePool with name {}".format(name))
[...]

PR #668, opened 2017-07-10
Build grafana dashboard for storage clusters

khaledkbadr:
relates to #634

Auto generating dashboard for storage clusters while deploying `statsdb` or `storage_cluster` services.

### DIFF
--- a/templates/storage_cluster/actions.py
+++ b/templates/storage_cluster/actions.py
@@ -71,7 +71,8 @@ def init(job):
             'devices': diskmap
         }
         storagepoolname = 'cluster_{}_{}_{}'.format(node.name, service.name, disk.name)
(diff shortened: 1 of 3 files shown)


templates/storage_cluster/actions.py
```diff
@@ -120,6 +121,16 @@ def create_server(node, disk, baseport, tcp, variant='data'):
         service.consume(storageEngine)
         service.model.data.storageEngines[index] = storageEngine.name
 
+    grafanasrv = service.aysrepo.serviceGet(role='grafana', instance='statsdb', die=False)
```
grimpy: when grafana is not installed this will crash with a notfound error
```
NotFound: ERROR: Cannot find service grerger:grger ((type:notfound)
```

PR #690, opened 2017-07-11
Graphs and Dashborads using Grafana

mie00:


api/node/node_api_CreateDashboard.go
```diff
@@ -0,0 +1,60 @@
...
+
+// CreateDashboard is the handler for POST /graph/{graphid}/dashboards
+// Creates a new dashboard
+func (api NodeAPI) CreateDashboard(w http.ResponseWriter, r *http.Request) {
+	aysClient := tools.GetAysConnection(r, api)
+	var reqBody Dashboard
+	vars := mux.Vars(r)
+	graphId := vars["graphid"]
+
+	if err := json.NewDecoder(r.Body).Decode(&reqBody); err != nil {
```
grimpy: this should check if dahsboard with reqBody.name exists and throw a StatusConflict incase it does

PR #697, opened 2017-07-13
update setup doc

yveskerwyn:
it was behind to some changes done in the alpha-4 documentation


docs/setup/README.md
```diff
@@ -4,21 +4,37 @@ This is the recommended and currently the only supported option to setup a Zero-
...
+7. [Setup the AYS Bootstrap service](#start-the-bootstrap-service)
+8. [Boot your Zero-OS nodes](#boot-your-zero-os-nodes)
+9. [Setup Statistics Monitoring](#setup-statistics-monitoring)
+
 
 ## Create a JumpScale9 Docker container
 
 Create the Docker container with JumpScale9 development environment by following the documentation at https://github.com/Jumpscale/developer#jumpscale-9.
-> **Important:** Make sure you set the `GIGBRANCH` environment variable to 9.0.0 before running `jsinit.sh`. This version of 0-orchestrator will only work with this version of JumpScale.
+> **Important:** Make sure you set the `GIGBRANCH` environment variable to **9.0.3** before running `jsinit.sh`. This version of 0-orchestrator will only work with this version of JumpScale.
```
zaibon: This is not the correct version of jumpscale for this branch. We need to wait for the new release of jumpscale to put the correct version here

yveskerwyn: Sure, but this should not block me PR, and 9.0.3 is still better than what was specified before, 9.0.0

So thanks for unblocking this PR, since there is more then just this change

Issue #680, opened 2017-07-11
Create AYS template for etcd cluster


zaibon:
### Goal: 
Beeing able to deploy a cluster of etcd on top of zero-os

### Requirements:
Cluster size can be 1, 3 or 5.

khaledkbadr:
Is the etcd cluster like the storage cluster but with etcd instead of ardb?

zaibon:
Not really,  etcd is a distributed key-value store not like ardb.
So in order to create a cluster of etcd you have to give the address of the other instance when starting an etcd. 
See for more detail : https://github.com/coreos/etcd/blob/master/Documentation/op-guide/clustering.md#static

khaledkbadr:
For cluster size, will the size be derived based on the number of nodes found or it should be specified through the blueprint?

khaledkbadr:
I add an initial implementation for the `etcd_cluster` service and he approach taken in it is

* Like `storage_cluster` nodes should be passed through blueprint to the `etcd_cluster` service
* Deployment of one etcd server per node. Therefore, size < #nodes is not allowed
* If #nodes is larger than the size, random sample of nodes would be chosen and etcd servers would be deployed to those nodes.

zaibon:
>Like storage_cluster nodes should be passed through blueprint to the etcd_cluster service 

perfect

> Deployment of one etcd server per node. Therefore, size < #nodes is not allowed

Make sure you only allow odd number of node in the cluste, otherwise cluster can't take decision

PR #714, opened 2017-07-16
Intial implementation of etcd cluster service

khaledkbadr:

relates to #680 

The approach taken here is 
- Like nodes should be passed through blueprint to the `etcd_cluster` service
- Deployment of one `etcd` server per node. Therefore, `size < #nodes` is not allowed
- If #nodes is larger than the size, random size number of nodes would be chosen and `etcd` servers would be deployed to those nodes.

### DIFF
--- a/templates/etcd_cluster/actions.py
+++ b/templates/etcd_cluster/actions.py
@@ -52,7 +52,12 @@ def init(job):
         etcd_service.consume(tcps[0])
         etcd_service.consume(tcps[1])
         service.consume(etcd_service)
+
+
+def install(job):
+    service = job.service
     service.model.data.status = "running"
+    service.saveAll()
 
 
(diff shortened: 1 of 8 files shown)


templates/etcd_cluster/actions.py
```diff
@@ -0,0 +1,81 @@
...
+
+    for k, v in etcd_args.items():
+        tcps = v.pop("tcps")
+        etcdname = "{}_{}_{}".format(service.name, k, tcps[1].model.data.port)
+        v["peers"] = peers
+        etcd_service = etcd_actor.serviceCreate(instance=etcdname, args=v)
+        etcd_service.consume(tcps[0])
+        etcd_service.consume(tcps[1])
+        service.consume(etcd_service)
+    service.model.data.status = "running"
```
mie00: Maybe better in install?

PR #716, opened 2017-07-17
bootstrap update for Threefold nodes

jvwilder:


templates/bootstrap.zero-os/actions.py
```diff
@@ -83,6 +83,22 @@ def try_authorize(job, logger, netid, member, zerotier):
         member = resp.json()
     zerotier_ip = member['config']['ipAssignments'][0]
 
+    # execute hardwarecheck if necessary
+    hwc_producers=service.producers.get('hardwarecheck')
+    if hwc_producers != None:
```
zaibon: for boolean condition use `is not` intead of `!=`

templates/bootstrap.zero-os/actions.py
```diff
@@ -83,6 +83,22 @@ def try_authorize(job, logger, netid, member, zerotier):
...
+        logger.info("Hardware check necessary")
+        hwcheck_job=hwc_producers[0].getJob('check', args={'ipaddr': zerotier_ip,
```
FastGeert: If get producers returns a list, we should loop them

templates/bootstrap.zero-os/schema.capnp
```diff
@@ -3,7 +3,9 @@
 struct Schema {
     zerotierNetID @0 :Text;
     zerotierToken @1 :Text;
+    # networks the new node needs to consume
```
zaibon: Need to update the schema ID when modifying it's content.

templates/bootstrap.zero-os/actions.py
```diff
@@ -83,6 +83,22 @@ def try_authorize(job, logger, netid, member, zerotier):
...
+                                                       'node_id': member['nodeId'],
+                                                       'jwt': get_jwt_token(service.aysrepo)})
+        j.tools.async.wrappers.sync(hwcheck_job.execute())
+
+    # execute odoo registration if necessary
+    odoo_producers=service.producers.get('odooregistration')
```
FastGeert: A more generic name will be better. Eg erp_registration.

An implementation for odoo can then be: erp_registration.odoo

FastGeert: Also erp registration should be moved to the complete bottom of the method.

Issue #666, opened 2017-07-10
implement error/stdout feedback mechanism to AYS


despiegk:

Dinaamagdy:
ays throw this error in watchdog in execution  process like creation of vm  ,then it doesn't work 

```
[Mon31 11:41] - Job.py              :56  :j.core.jobcontroller.job.e73be49105ca2b785132c60538ee94bd - ERROR    - Traceback (most recent call last):
  File "/usr/lib/python3.5/asyncio/tasks.py", line 241, in _step
    result = coro.throw(exc)
  File "/tmp/actions/node.zero-os/f3f606e0fa7524d9dc4ecdf4b22032b1.py", line 391, in streaming
    queue = await cl.subscribe('ays.monitor')
  File "/usr/local/lib/python3.5/dist-packages/zeroos/orchestrator/sal/Pubsub.py", line 137, in subscribe
    response = await self.sync('logger.subscribe', {'queue': queue})
  File "/usr/local/lib/python3.5/dist-packages/zeroos/orchestrator/sal/Pubsub.py", line 125, in sync
    response = await self.raw(command, args)
  File "/usr/local/lib/python3.5/dist-packages/zeroos/orchestrator/sal/Pubsub.py", line 114, in raw
    self._redis = await self.get()
  File "/usr/local/lib/python3.5/dist-packages/zeroos/orchestrator/sal/Pubsub.py", line 73, in get
    timeout=self.timeout)
  File "/usr/local/lib/python3.5/dist-packages/aioredis/commands/__init__.py", line 148, in create_redis
    loop=loop)
  File "/usr/local/lib/python3.5/dist-packages/aioredis/connection.py", line 77, in create_connection
    host, port, ssl=ssl, loop=loop), timeout, loop=loop)
  File "/usr/lib/python3.5/asyncio/tasks.py", line 385, in wait_for
    yield from waiter
  File "uvloop/future.pyx", line 230, in __iter__ (uvloop/loop.c:110600)
RuntimeError: Task <Task pending coro=<watchdog.<locals>.streaming() running at /tmp/actions/node.zero-os/f3f606e0fa7524d9dc4ecdf4b22032b1.py:391> cb=[_execute_cb(job: node.zer...340 (watchdog))() at /opt/code/github/jumpscale/ays$/JumpScale9AYS/jobcontroller/Job.py:18, Task._wakeup()]> got Future <Future pending> attached to a different loop
[...]

Issue #489, opened 2017-06-19
Self-healing : NBD server failure


zaibon:
Specs : https://github.com/zero-os/0-orchestrator/blob/master/specs/selfhealing/storage.md

Dinaamagdy:
- after  killing  nbd server process, Ays still has no failure in nbd server 
## Software Version
967b777e18d52e8ce56737d5ac7ddf5682c65778

ahmedelsayed-93:
issue still exists
### software version
master : 2043f538

Dinaamagdy:
blocked due to https://github.com/zero-os/0-orchestrator/issues/666

Dinaamagdy:
it doesn't work after migration 
## Steps
- create vm in node1
- migrate vm to node2
- kill nbdeserver for this vm in node2
## Error
```
traceback (most recent call last):
  File "/usr/lib/python3.5/concurrent/futures/thread.py", line 55, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/tmp/actions/nbdserver/ed1ccdc6bb5a843ef31d39df3c950018.py", line 195, in watchdog_handler
    vm_service = service.aysrepo.serviceGet(role='vm', instance=service.name)
  File "/opt/code/github/jumpscale/ays9/JumpScale9AYS/ays/lib/AtYourServiceRepo.py", line 349, in serviceGet
    (role, instance), level=1, source="", tags="", msgpub="")
JumpScale9.errorhandling.JSExceptions.NotFound: ERROR: Cannot find service vm:vdisks_vm2_248a07e3cc20 ((type:notfound)
[Mon07 08:52] - Job.py              :56  :j.core.jobcontroller.job.b0cd3dd4b8302ebe533c8ff55dba58cc - ERROR    - Traceback (most recent call last):
  File "uvloop/future.pyx", line 374, in uvloop.loop.BaseTask._fast_step (uvloop/loop.c:112704)
  File "/tmp/actions/node.zero-os/aac16753b7bb0b6f5c7f9f7004b95aef.py", line 233, in streaming
    await cl.global_stream(queue)
  File "/usr/local/lib/python3.5/dist-packages/zeroos/orchestrator/sal/Pubsub.py", line 97, in global_stream
    await self.callback(job_id, meta >> 16, line, meta & 0xff)
  File "/tmp/actions/node.zero-os/aac16753b7bb0b6f5c7f9f7004b95aef.py", line 215, in callback
[...]

PR #739, opened 2017-07-20
Add a watchdog action to the node service

khaledkbadr:
Add a watchdog action to the node service
- it's purpose is to listen on the logs of that node 
- if interested log is found, the log details is passed to an action defined on the service related `watchdog_handler`. So by that design each service handle its logs accordingly 

Conventions:
- The `watchdog_handler` action name is a convention.
- Processes that needs to be watched should have an id of `role.service_name` and must be added to the `watched_roles` dict in the `watchdog` action in the nodes service

related to: #489 and #666


templates/node.zero-os/actions.py
```diff
@@ -202,3 +201,66 @@ def uninstall(job):
...
+    async def callback(jobid, level, message, flag):
+        if "." not in jobid:
+            return
+        role, instance = jobid.split(".", 1)
+        if role not in watched_roles:
+            return
+
+        eof = flag & 0x6 != 0
+        matched_messages = watched_roles[role].get("message", None)
+        if not (matched_messages and message in matched_messages) and not (watched_roles[role]["eof"] and eof):
```
FastGeert:
```python
matcher = watched_roles[role].get("message", None)
if not (matcher and matcher.match(message) and not (watched_roles[role]["eof"] and eof):
```

PR #759, opened 2017-07-26
Add temperature healthcheck

khaledkbadr:


pyclient/zeroos/orchestrator/sal/healthchecks/temperature.py
```diff
@@ -0,0 +1,68 @@
...
+        "name": resource,
+        "category": "Hardware",
+        "id": "Temperature",
+    }
+    if status != "OK":
+        return result
+
+    if temperature >= WARNING_TRIPPOINT:
+        result["status"] = "WARNING"
+        result["message"] = "%s's average temperature is too high" % (resource)
```
grimpy: you remove the original message which contains the actual temperatue which is not nice.
i dont think you need to bother with changing the message. i think it would be enough by changing the status

pyclient/zeroos/orchestrator/sal/healthchecks/temperature.py
```diff
@@ -0,0 +1,68 @@
...
+    elif temperature >= ERROR_TRIPPOINT:
+        result["status"] = "ERROR"
+        result["message"] = "%s's average temperature is too high" % (resource)
```
grimpy: same as warning

PR #772, opened 2017-07-27
Interrupts health check

muhamadazmy:


pyclient/zeroos/orchestrator/sal/healthchecks/interrupts.py
```diff
@@ -0,0 +1,64 @@
...
+class Interrupts(HealthCheckRun):
+    ID = 'interrupts'
+
+    def __init__(self, warn=8000, error=1000):
+        super().__init__()
+        self._warn = warn
+        self._error = error
+        self.result['id'] = 'interrupts'
+        self.result['name'] = 'CPU Interrupts'
+        self.result['category'] = 'Hardware'
```
grimpy: resource isnt filled other then this it looks good,
Also if you rebasse you will notice i have changed all healtchecks to pass node inside init and store it under self.node so the resource can be filed in the __init__

PR #776, opened 2017-07-27
Adds power supply health check

ashraffouda:


pyclient/zeroos/orchestrator/sal/healthcheck.py
```diff
@@ -120,3 +120,9 @@ def check_interrupts(self):
         from .healthchecks.interrupts import Interrupts
         inter = Interrupts(self.node)
         return inter.start()
+
+    def powersupply(self, container):
+        from .healthchecks.powersupply import PowerSupply
+        self.node.client.bash("modprobe ipmi_si && modprobe ipmi_devintf").get()
```
grimpy: I'm thinking since we need this in multiple healthchecks that we add this in the node.install to only run this once instead of every 30seconds per healtcheck

PR #779, opened 2017-07-27
Make sure params of healtcheck are correct type

grimpy:

### DIFF
--- a/pyclient/zeroos/orchestrator/sal/healthcheck.py
+++ b/pyclient/zeroos/orchestrator/sal/healthcheck.py
@@ -12,7 +12,7 @@ class HealthCheckRun:
         self.category = category
         self._messages = []
         self.resource = resource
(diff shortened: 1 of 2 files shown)


pyclient/zeroos/orchestrator/sal/healthcheck.py
```diff
@@ -25,7 +25,7 @@ def start(self, *args, **kwargs):
             'name': self.name,
             'category': self.category,
             'resource': self.resource,
-            'stacktrace': self.stacktrace,
+            'stacktrace': self.stacktrace or '',
```
serboctor: why is this or needed? you already initialize it with ''.

grimpy: its possible it gets set wrong from the run itself

PR #796, opened 2017-07-31
Persistent zerotierid for container

jvwilder:


templates/container/actions.py
```diff
@@ -78,6 +78,9 @@ def start(job):
         if nic.type == 'zerotier':
             zerotier_nic_config(service, job.logger, container, nic)
 
+    if not service.model.data.identity:
+        service.model.data.identity = container.client.zerotier.info()['secretIdentity']
```
grimpy: you can only do this if you have atleast one zerotier nic otherwise zerotier.info will return an error as there is no identity set.

PR #802, opened 2017-07-31
Gateway watchdog

serboctor:


pyclient/zeroos/orchestrator/sal/gateway/dhcp.py
```diff
@@ -29,7 +29,7 @@ def apply_config(self):
                     time.sleep(0.2)
                 break
 
-        self.container.client.system(DNSMASQ)
+        self.container.client.system(DNSMASQ, id='dhcp.{}'.format(self.container.name))
```
khaledkbadr: why not use the role of the service `service.role` to be consistent

templates/http/actions.py
```diff
@@ -26,3 +26,9 @@ def apply_rules(job, httpproxies=None):
 
 def update(job):
     apply_rules(job, job.model.args["httpproxies"])
+
+
+def watchdog_handler(job):
+    gateway = job.service.parent.consumers['gateway'][0]
+    if gateway.model.data.status == 'running':
+        start(job)
```
khaledkbadr: use ```asyncio.ensure_future(service.executeAction('stop'))``` in order not to block the event loop. Check the nbdserver watchdog_handler for reference

pyclient/zeroos/orchestrator/sal/gateway/dhcp.py
serboctor: This is in the sal...there is no service.

Issue #618, opened 2017-07-04
Create selfhealingscript to cleanup ssh daemons created by migrations


grimpy:
Check if there are still ssh daemons running on the node which are left over from vm migrations

despiegk:
lets discuss how this will be implemented

abdulrahmantkhalifa:
things that will be cleaned up 
 - ssh.config files that realate to **finished** or **errored** migrations located in /tmp 
 - sshd processes that relates to **finished** or **errored** migrations 
 - tcp services with insance name tcp_node_port that realtes to **finished** or **errored** migrations
 - opened ports for  **finshed** or **errored** migration

Dinaamagdy:
tcp service still exist after migration end 
```
---------------------------------------------------                     [26/768]
Service: tcp_248a07e76f80_4000 - Role: tcp
state : ok
key : f56191e751004185b0e64eb56f8e57be

Instance data:
- node : 248a07e76f80
- port : 4000

Parent:
node!248a07e76f80

Children: None

Producers:
node!248a07e76f80

Consumers: None

Recurring actions: None

Event filters: None
```
## Software Version
5cfc4d99b8c58fce0596a1e9551e1a2fefcac1d7

abdulrahmantkhalifa:
cannot reproduce

ahmedelsayed-93:
ssh.config still exists on the new node after migrating 

![image](https://user-images.githubusercontent.com/24386820/29418730-70a988d8-836d-11e7-8275-f11b08e7a631.png)

### software versions
master : 96700d6

abdulrahmantkhalifa:
cannot reproduce

Dinaamagdy:
still sshd in new node exist 
as shown 
```
STATE: 0 SUCCESS
STDOUT:
 1222 root       0:00 /usr/sbin/sshd -e -D
 7091 root       0:00 /usr/sbin/sshd -f /tmp/ssh.config_vm6_4000
 7201 root       0:00 /usr/sbin/sshd -f /tmp/ssh.config_vm6_4001
21659 root       0:00 /bin/sh -c ps -a |grep sshd
21662 root       0:00 grep sshd

STDERR:

DATA:
```
## Software Version 
e7383bcca187748a6ac6f7e54feb4fa5433e843b

Dinaamagdy:
## Software Version 
5d71c78faa37c0316e4d5f2143e9e6096d8f7fa9

PR #804, opened 2017-07-31
Added new self healing healthcheck for ssh deamon created in migrate

abdulrahmantkhalifa:
Adjusted code to allow the script to work easier
relates to issue #618

### DIFF
--- a/templates/vm/actions.py
+++ b/templates/vm/actions.py
@@ -22,7 +22,6 @@ def get_node(job):
     from zeroos.orchestrator.sal.Node import Node
     return Node.from_ays(job.service.parent, job.context['token'])
 
-
 def create_zerodisk_container(job, parent):
     """
     first check if the vdisks container for this vm exists.
@@ -88,7 +87,7 @@ def _nbd_url(job, container, nbdserver, vdisk):
     from zeroos.orchestrator.sal.Node import Node
 
     container_root = container.info['container']['root']
-    node = Node.from_ays(nbdserver.parent.parent, password=job.context['token'])._client
+    node = Node.from_ays(nbdserver.parent.parent, password=job.context['token']).client
     node.filesystem.mkdir("/var/run/nbd-servers/")
     endpoint = nbdserver.model.data.socketPath.lstrip('/')
     socket_path = j.sal.fs.joinPaths(container_root, endpoint)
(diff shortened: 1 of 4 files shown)


templates/vm/actions.py
```diff
@@ -504,14 +500,20 @@ def migrate(job):
     nbdserver.consume(vdisk_container)
     service.consume(nbdserver)
     service.consume(vdisk_container)
-    target_node_client = Node.from_ays(target_node, job.context['token'])._client
-    node_client = Node.from_ays(service.parent, job.context['token'])._client
+    target_node_client = Node.from_ays(target_node, job.context.get('token',''))._client
```
grimpy: why use the attribute instead of the property?

templates/vm/actions.py
```diff
@@ -88,7 +87,7 @@ def _nbd_url(job, container, nbdserver, vdisk):
     from zeroos.orchestrator.sal.Node import Node
 
     container_root = container.info['container']['root']
-    node = Node.from_ays(nbdserver.parent.parent, password=job.context['token'])._client
+    node = Node.from_ays(nbdserver.parent.parent, password=job.context.get('token', ""))._client
```
grimpy: incosistante usage of get all other places just do .get instead of passing empty string

templates/vm/actions.py
```diff
@@ -417,28 +409,36 @@ def start_migartion_channel(job, old_node, node):
...
-
+        file_discriptor = node.client.filesystem.open("/root/.ssh/authorized_keys", mode='a')
         node.client.filesystem.write(file_discriptor, pub_key)
         old_node.client.filesystem.close(file_discriptor)
-        old_node.client.bash('ssh-keyscan %s >> /root/.ssh/known_hosts' % node.addr).get()
 
+        # write the ssh identy into the known hosts of old node if it doesnt exist
+        ssh_identity = old_node.client.system('ssh-keyscan -p %s %s' % (port, node.addr)).get().stdout
+        if not ssh_identity:
+            raise j.exceptions.RuntimeError('could not generate the ssh identity')
```
grimpy: you mean could not fetch the ssh idenity

templates/vm/actions.py
```diff
@@ -417,28 +409,36 @@ def start_migartion_channel(job, old_node, node):
...
+        exists = old_node.client.filesystem.exists("/root/.ssh/known_hosts")
+        if exists:
+            file_discripter = old_node.client.filesystem.open("/root/.ssh/known_hosts", mode='r')
+            known_hosts = node.client.filesystem.read(file_discripter)
+            old_node.client.filesystem.close(file_discripter)
+        if not exists or ssh_identity not in known_hosts:
+            file_discripter = old_node.client.filesystem.open("/root/.ssh/known_hosts", mode='a')
```
grimpy: this should probably be a+ to allow creation if the file does not exists

Issue #580, opened 2017-07-03
Monitor ARDB server process health


FastGeert:
Currently we only check if the process is running. Thats not good enough. We also need to investigate the healthyness of the process.

So we need to execute the following to validate if the ARDB is still healthy:
```
127.0.0.1:16379> SET testkey geertwashere
OK
127.0.0.1:16379> GET testkey
"geertwashere"
127.0.0.1:16379> DEL testkey
(integer) 1
127.0.0.1:16379> EXISTS testkey
(integer) 0
```

khaledkbadr:
To implement ardb healthcheck we need `nbdserver` to add support for retrying to connect to the ardb server which is not implemented yet so this issue is blocked until such support exist

grimpy:
We need to change ardb to listen on the storage network and the management network, then we can in the monitor (recuring task) of the ardb service implement this check which can change the status of the service

Dinaamagdy:
can't validate

PR #813, opened 2017-08-02
Add ardb health check

khaledkbadr:
related to #580 

### DIFF
--- a/templates/storage_engine/actions.py
+++ b/templates/storage_engine/actions.py
@@ -11,6 +11,7 @@ def start(job):
     service = job.service
     storageEngine = StorageEngine.from_ays(service, job.context['token'])
     storageEngine.start()
+    service.model.data.status = 'running'
 
 
 def stop(job):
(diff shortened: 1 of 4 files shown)


templates/storage_engine/actions.py
```diff
@@ -39,6 +41,11 @@ def monitor(job):
...
                 if running:
                     service.model.dbobj.state = 'ok'
+                    service.model.data.status = 'running'
             except:
                 job.logger.error("can't restart storageEngine {} not running".format(service.name))
                 service.model.dbobj.state = 'error'
+                service.model.data.status = 'halted'
+        else:
+            if not storageEngine.is_healthy():
+                service.model.data.status = "unhealthy"
```
grimpy: you never put status back to running in cause it return true again.

templates/storage_engine/actions.py
```diff
@@ -39,6 +41,11 @@ def monitor(job):
                 running, _ = storageEngine.is_running()
                 if running:
                     service.model.dbobj.state = 'ok'
+                    service.model.data.status = 'running'
```
grimpy: in this case the is_healthy check never gets executed.

PR #827, opened 2017-08-04
Fix vm delete ncase tlog and nbdserver are on same node

grimpy:
Collect all parent of nbd and tlogserver and delete after deletion of
nbd/tlog


xmonader:
Nice catch!

PR #829, opened 2017-08-06
changed behaviour of process change in vm actions

abdulrahmantkhalifa:
if a parameter such as vdisk or nic is passed as empty it will remove the existing ones , if not passed at all in the blueprint it will remain
the same.

### DIFF
--- a/templates/vm/actions.py
+++ b/templates/vm/actions.py
@@ -493,9 +493,6 @@ def migrate(job):
 
     # start migration channel to copy keys and start ssh deamon
     ssh_port, job_id = start_migartion_channel(job, old_node, target_node)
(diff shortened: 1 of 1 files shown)


templates/vm/actions.py
```diff
@@ -570,30 +557,25 @@ def _diff(col1, col2):
 
 
 def updateDisks(job, client, args):
+    if args.get('disks') == None:
```
grimpy: is None

templates/vm/actions.py
```diff
@@ -626,17 +608,19 @@ def updateDisks(job, client, args):
 
 
 def updateNics(job, client, args):
+    if args.get('disks') == None:
```
grimpy: this should be nics

Issue #661, opened 2017-07-09
Extend grafana service


serboctor:
Currecntly the grafana service doesn't support the list on influxdbs in the influxdb argument in the schema. The processChange should be extended to handle adding/removing influxdbs.

Dinaamagdy:
can't validate now as it depends on new changes on js9 not included in 9.1.1 release

Dinaamagdy:
- can't extend grafana  with another influxdb 
## Steps 
- create two influxdb 
- create grafana with one of them
-  check that data source exist 
-  then try to add another one 
## Software Version 
6bb174edd298452e053fb8d915194b96c5c44a92

rkhamis:
@serboctor @zaibon is there a practical use case for this FR? According to @khaledkbadr there isn't even API exposing adding influxdb and grafana

rkhamis:
Will put this on roadmap since no one seems to have any usecase for it

PR #871, opened 2017-08-13
grafana: support influxdb in process change

mie00:
See #661

### DIFF
--- a/templates/grafana/actions.py
+++ b/templates/grafana/actions.py
@@ -11,12 +11,16 @@ def get_container(service, force=True):
     return containers[0]
 
 
-def configure_datasources(job, grafana):
-    service = job.service
-    influxes = service.producers.get('influxdb')
+def add_datasources(grafana, influxes):
     for influx in influxes:
         for i, database in enumerate(influx.model.data.databases):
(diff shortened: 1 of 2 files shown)


templates/grafana/actions.py
```diff
@@ -92,26 +96,46 @@ def processChange(job):
...
-    service.model.data.port = args['port']
-
-    # @TODO: Handle influxdb list change
+    elif args.get('influxdb'):
+        service.model.data.influxdb = args['influxdb']
+        added = []
+        removed = []
+        new = set(args['influxdb'])
+        old = set(s.name for s in service.producers.get('influxdb'))
+        print(new, old)
```
serboctor: Forgot to remove it?

templates/grafana/actions.py
```diff
@@ -11,14 +11,18 @@ def get_container(service, force=True):
     return containers[0]
 
 
-def configure_datasources(job, grafana):
-    service = job.service
-    influxes = service.producers.get('influxdb')
+def configure_datasources(grafana, influxes):
```
serboctor: Since we now have a "remove_datasources" function then may be we should rename this to "add_datasources" so it wouldn't be confusing for everyone else?

Issue #765, opened 2017-07-26
Rest api: update container nics with put request


jvwilder:

zaibon:
AYS service has the support for update but we need to expose this feature over the REST API

abdulrahmantkhalifa:
Only NIC changes are implemented in the process change of the container actions, should I implement other data changes if so, which.

Also concerning this  `AYS service has the support for update but we need to expose this feature over the REST API` isn't it already exposed as execute blueprint  isn't it enough to execute a blueprint with new data to call the process change action, or there other functionality missing .

muhamadazmy:
Actually the only "dynamic" feature of a container that u can update on the fly is NICS, by adding and removing nics. Name, tags, mounts and others are not changeable once the container is started

abdulrahmantkhalifa:
got it thank you

Dinaamagdy:
## Steps 
- make sure that ovc configuration doesn't exist 
- create container without nic 
- use put API to put vlan nic 

## Issue 
it gives right  error in ays  but it give 201 in put request and vlan network added when u get this container: 
```
INFO:network:
[Wed16 14:18] - Job.py              :56  :j.core.jobcontroller.job.64ab5fa3a5c9fc2147491cf751057e2d - ERROR    - Traceback (most recent call last):
  File "/usr/lib/python3.5/concurrent/futures/thread.py", line 55, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/tmp/actions/container/5db963e8dc61f38fa3bab5f9bf22429d.py", line 275, in processChange
    update(service, job.logger, job.context['token'], args['nics'])
  File "/tmp/actions/container/5db963e8dc61f38fa3bab5f9bf22429d.py", line 250, in update
    cl.nic_add(container.id, nic_dict)
  File "/usr/local/lib/python3.5/dist-packages/zeroos/core0/client/client.py", line 1120, in nic_add
    return self._client.json('corex.nic-add', args)
  File "/usr/local/lib/python3.5/dist-packages/zeroos/core0/client/client.py", line 726, in json
    result = self.sync(command, arguments, tags=tags, id=id)
  File "/usr/local/lib/python3.5/dist-packages/zeroos/core0/client/client.py", line 716, in sync
    raise RuntimeError('invalid response: %s' % result.state, result)
RuntimeError: ('invalid response: ERROR', STATE: ERROR
STDOUT:

STDERR:

DATA:
"ovs is needed for VLAN network type"
)
[...]

PR #874, opened 2017-08-13
added api for container update

abdulrahmantkhalifa:
only available update through the client are the nics so that is what
was added,
relates to issue #765

### DIFF
--- a/api/node/node_api_UpdateContainer.go
+++ b/api/node/node_api_UpdateContainer.go
@@ -1,0 +1,65 @@
+package node
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/gorilla/mux"
+
+	"github.com/zero-os/0-orchestrator/api/tools"
+)
+
+// UpdateContainer is the handler for PUT /nodes/{nodeid}/containers/{containername}
+// Update a new Container
+func (api NodeAPI) UpdateContainer(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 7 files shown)


api/node/node_api_UpdateContainer.go
```diff
@@ -0,0 +1,65 @@
...
+	containerName := vars["containername"]
+
+	// validate container name
+	exists, err := aysClient.ServiceExists("container", containerName, api.AysRepo)
+	if err != nil {
+		tools.WriteError(w, http.StatusInternalServerError, err, "Error checking container service exists")
+		return
+	} else if !exists {
+		err = fmt.Errorf("Container with name %s does not exists", containerName)
+		tools.WriteError(w, http.StatusConflict, err, "")
```
serboctor: Shouldn't this be not found?

api/node/node_api_UpdateContainer.go
```diff
@@ -0,0 +1,65 @@
...
+			return
+		}
+	}
+
+	obj := make(map[string]interface{})
+	obj[fmt.Sprintf("container__%s", containerName)] = reqBody
+
+	if _, err := aysClient.UpdateBlueprint(api.AysRepo, "container", containerName, "install", obj); err != nil {
+		errmsg := fmt.Sprintf("error executing blueprint for container %s creation", containerName)
+		tools.WriteError(w, http.StatusConflict, err, errmsg)
```
serboctor: Why is this a statusconflict?

PR #883, opened 2017-08-15
Fixes disk rollback

ashraffouda:


templates/vdisk/actions.py
```diff
@@ -148,6 +138,17 @@ def save_config(job):
         if result.state != "SUCCESS":
             raise RuntimeError("Failed to save tlog conf storage: %s" % service.name)
 
+    # push nbd config to etcd
+    config = {
+        "storageClusterID": service.model.data.storageCluster,
+        "templateStorageClusterID": templateStorageclusterId,
```
khaledkbadr: Is templateStorageClusterID default to empty string if template not found?

templates/vdisk/actions.py
```diff
@@ -225,10 +226,14 @@ def rollback(job):
     from zeroos.orchestrator.sal.ETCD import EtcdCluster
 
     service = job.service
+    if 'vm' not in service.consumers:
+        raise j.exceptions.Input('Can not rollback a disk that is not attached to a vm')
     service.model.data.status = 'rollingback'
     ts = job.model.args['timestamp']
-
     tlogclusterconfig = get_cluster_config(job, type='tlog')
+    if not tlogclusterconfig:
```
khaledkbadr: Can you add that check to the API as well?

ashraffouda: this is not a bad request this should be an error 
cause tlogserver should be their anyway

khaledkbadr: Not always, you can always create a vdisk without a tlogcluster

templates/vdisk/actions.py
ashraffouda: yeah, see the full method

templates/vdisk/actions.py
ashraffouda: you shouldn't enter if, unless there is unexpected behavior in the env

PR #887, opened 2017-08-16
Fixes getting node from ays with jwt

ashraffouda:

### DIFF
--- a/templates/container/actions.py
+++ b/templates/container/actions.py
@@ -22,14 +22,16 @@ def input(job):
 
 def init(job):
     from zeroos.orchestrator.sal.Node import Node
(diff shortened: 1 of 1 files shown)


templates/container/actions.py
```diff
@@ -22,14 +22,16 @@ def input(job):
 
 def init(job):
     from zeroos.orchestrator.sal.Node import Node
+    from zeroos.orchestrator.configuration import get_jwt_token
     service = job.service
+    jwt = get_jwt_token(service.aysrepo)
```
abdulrahmantkhalifa: set the job.context['token'] to the jwt so other places can use it

PR #924, opened 2017-08-29
Check if iyo configuration is valid when deploying object cluster

khaledkbadr:

### DIFF
--- a/templates/storage_cluster/actions.py
+++ b/templates/storage_cluster/actions.py
@@ -235,6 +235,7 @@ def init(job):
 
 def save_config(job):
     import yaml
+    import requests
     from zeroos.orchestrator.sal.StorageCluster import StorageCluster
     from zeroos.orchestrator.sal.ETCD import EtcdCluster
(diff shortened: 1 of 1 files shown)


templates/storage_cluster/actions.py
```diff
@@ -255,12 +256,32 @@ def save_config(job):
...
+    url += "&grant_type=client_credentials&response_type=id_token&scope=user%3Amemberof%3A{org}.0stor.{namespace}.read%2"
+    url += "Cuser%3Amemberof%3A{org}.0stor.{namespace}.write%2Cuser%3Amemberof%3A{org}.0stor.{namespace}.delete"
+
+    url = url.format(
+        client_id=iyo_client_id,
+        client_secret=iyo_secret,
+        org=iyo_org,
+        namespace=iyo_namespace
+    )
+    res = requests.post(url)
```
grimpy: i dont like the way the questy params are contructed, request.post as a parameters called params in which you can pass a dict which are the quiery params, also since its post it should probably be passed in the body instead.

PR #953, opened 2017-09-04
fix test096

Dinaamagdy:


tests/0_orchestrator/test_suite/testcases/extend_tests/test01_containers_apis.py
```diff
@@ -650,7 +650,7 @@ def test015_create_containers_with_open_ports(self):
         ports = "%i:%i" % (hostport, containerport)
         nics = [{"type": "default"}]
 
-        # create rule on port 7070
+        self.lg.info("[*] Create rule on port 7070")
         try:
             self.core0_client.client.nft.open_port(hostport)
         except:
```
FastGeert: An empty expect clause without logging? We should at least log the exception, right?

Dinaamagdy: yess, right ,we can add logs here

PR #977, opened 2017-09-06
Container backup

muhamadazmy:
- Introducing the /backup end point to create container backups
- container create can now accept a 'restic' url to create a container from a backup.


api/backup/BackupContainer.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"fmt"
+	"net/url"
+)
+
+type BackupContainer struct {
+	Name      string `json:"name"`
+	Container string `json:"container"`
+	URL       string `json:"url"`
```
abdulrahmantkhalifa: validate:"nonzero"` this can be used for validation instead of the condition , but import validate:"nonzero"` and validate see https://github.com/zero-os/0-orchestrator/blob/master/api/node/CPUInfo.go for example

muhamadazmy: Updated

api/backup/backup_api_BackupContainer.go
```diff
@@ -0,0 +1,60 @@
+package backup
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/zero-os/0-orchestrator/api/tools"
+	"net/http"
+)
+
+// CreateContainer is the handler for POST /backup
```
serboctor: Fix docs

api/backup/backup_api_BackupContainer.go
```diff
@@ -0,0 +1,60 @@
...
+		tools.WriteError(w, http.StatusInternalServerError, err, "Error checking container service exists")
+		return
+	} else if !exists {
+		err = fmt.Errorf("Container with name %s does not exists", reqBody.Container)
+		tools.WriteError(w, http.StatusNotFound, err, "")
+		return
+	}
+
+	bp := map[string]interface{}{
+		fmt.Sprintf("container_backup__%s", reqBody.Name): map[string]interface{}{
```
serboctor: Don't you need to check if another backup with the same name exists before creating?

api/backup/backup_api_BackupList.go
```diff
@@ -0,0 +1,42 @@
...
+	"net/http"
+)
+
+type Backup struct {
+	Name     string `json:"name"`
+	Snaphost string `json:"snapshot"`
+	URL      string `json:"url"`
+}
+
+// CreateContainer is the handler for POST /backup
```
serboctor: fix docs

raml/api.raml
```diff
@@ -1259,6 +1259,21 @@ types:
             responses:
               204:
                 description: Virtual machine migration request passed successfully
+/backup:
```
serboctor: where is the Get part?

api/backup/backup_api_BackupContainer.go
muhamadazmy: Won't this fail in ays anyway ?

muhamadazmy: done above

api/backup/backup_api.go
```diff
@@ -0,0 +1,29 @@
+package backup
+
+import (
+	ays "github.com/zero-os/0-orchestrator/api/ays-client"
+	_ "github.com/zero-os/0-orchestrator/api/validators"
+)
+
+// BackupAPI is API implementation of /node root endpoint
```
serboctor: /node?

api/backup/backup_api_BackupContainer.go
serboctor: As far as I remember it doesnt return the correct error and that's why in other apis we check if the service exists before attempting to create it.

muhamadazmy: Well, that's not a very good solution because a race condition might cause another service to be created with the same name after the exists check pass

muhamadazmy:
@FastGeert I thought about this a lot and decided that the best place to put it is under /backup for the following reasons:

- The backup is totally independent from the container, once u take a backup u can delete the full container or the full node without losing the backup, and then use this backup to create a new container on a different node.
- This way u can list all the available backup objects that was ever taken for any container that ever existed.
- Once the backup is taken, it doesn't relate to the container anymore, there not even a reference to the container it was created from. Since backup are not used to restore to the original containers, instead they are used to create new containers as if it's an flist.
- Since container name is globally unique per repo, I don't need to specify the node which is hosting this container.
- Think about containers backups as a `docker commit` command which creates a new image out of a running docker container, after the image is taken the container can be deleted and then new containers are created from the 'backups'

FastGeert:
Okay, you reasoning makes sense. Still I would add all the metadata of the container to the backup (flist, network configuration, mounts, ...) so that we have something more than just a name. Also add the type of the backup (backup of a container), because in the next milestone we are also adding backup of virtual machines.

PR #980, opened 2017-09-06
Fix choosing random node if halted

khaledkbadr:

### DIFF
--- a/templates/vdisk/actions.py
+++ b/templates/vdisk/actions.py
@@ -188,7 +188,12 @@ def get_templatecluster(job):
     if template.scheme == 'ardb' and template.netloc:
         return template.netloc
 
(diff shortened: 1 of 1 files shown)


templates/vdisk/actions.py
```diff
@@ -188,7 +188,7 @@ def get_templatecluster(job):
     if template.scheme == 'ardb' and template.netloc:
         return template.netloc
 
-    node_srv = service.aysrepo.servicesFind(role='node')[0]
+    node_srv = [node for node in service.aysrepo.servicesFind(role="node") if node.model.data.status != "halted"][0]
```
muhamadazmy: This still can cause index out of range error
it's better if u check on the returned list length before u try to get of(0). Then raise a proper error message if no matches were found

PR #991, opened 2017-09-07
disaster recovery in etcdcluster if service fails

abdulrahmantkhalifa:


templates/etcd_cluster/actions.py
```diff
@@ -129,6 +129,7 @@ def watchdog_handler(job):
     etcds = set(service.producers.get('etcd', []))
     working_etcds = set()
     dead_nodes = set()
+    dead_etcds = set()
```
muhamadazmy: make sure u save the data to make sure other routines sees the correct status

Issue #976, opened 2017-09-06
migrate vm with invalid nodeid returns response code 204 instead of 400 


ahmedelsayed-93:
### API
```POST  /nodes/{nodeid}/vms/{vmid}/migrate```
### request & response 
![image](https://user-images.githubusercontent.com/24386820/30104194-2a84a774-92f5-11e7-9aa1-e8ed08b8cb18.png)

### ays log
```
2017-09-06 09:16:53 - (network)[INFO][127.0.0.1:33220]: GET http://127.0.0.1:5000/ays/repository/orchestrator-server/service/vm/10-vm  200 29656
INFO:network:
2017-09-06 09:16:56 - (network)[INFO][127.0.0.1:33220]: POST http://127.0.0.1:5000/ays/repository/orchestrator-server/blueprint  201 237
INFO:network:
init done
blueprint done
[Wed06 09:16] - Job.py              :56  :j.core.jobcontroller.job.d4c46e68953a9eab0bb9f1a6738261c2 - ERROR    - Traceback (most recent call last):
  File "/usr/lib/python3.5/concurrent/futures/thread.py", line 55, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/tmp/actions/vm/bee59edc0cc2d9e89acc1336879f7af3.py", line 28, in migrate
    target_node = service.aysrepo.serviceGet('node', node)
  File "/opt/code/github/jumpscale/ays9/JumpScale9AYS/ays/lib/AtYourServiceRepo.py", line 349, in serviceGet
    (role, instance), level=1, source="", tags="", msgpub="")
JumpScale9.errorhandling.JSExceptions.NotFound: ERROR: Cannot find service node:123456 ((type:notfound)
[Wed06 09:16] - AtYourServiceRepo.py:687 :j.atyourservice.server         - INFO     - AYS Repo orchestrator-server auto commit done
[Wed06 09:16] - AtYourServiceRepo.py:703 :j.atyourservice.server         - WARNING  - Auto Push skipped, please make sure that you are using ssh url as a remote to be able to auto push
2017-09-06 09:16:58 - (network)[INFO][127.0.0.1:33220]: POST http://127.0.0.1:5000/ays/repository/orchestrator-server/blueprint/vm_10-vm_migrate_1504689413  200 113
INFO:network:
[...]

PR #1000, opened 2017-09-10
handled error if node does not exists in live mig

abdulrahmantkhalifa:
handled error if node does not exist in live migration
relates to issue #976

### DIFF
--- a/api/node/node_api_MigrateVM.go
+++ b/api/node/node_api_MigrateVM.go
@@ -56,6 +56,16 @@ func (api NodeAPI) MigrateVM(w http.ResponseWriter, r *http.Request) {
 		Node: reqBody.Nodeid,
 	}
 
+	_, res, err = aysClient.Ays.GetServiceByName(reqBody.Nodeid, "node", api.AysRepo, nil, nil)
+	if res.StatusCode == http.StatusNotFound {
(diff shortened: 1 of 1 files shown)


api/node/node_api_MigrateVM.go
```diff
@@ -56,6 +56,13 @@ func (api NodeAPI) MigrateVM(w http.ResponseWriter, r *http.Request) {
 		Node: reqBody.Nodeid,
 	}
 
+	_, res, _ = aysClient.Ays.GetServiceByName(reqBody.Nodeid, "node", api.AysRepo, nil, nil)
+	if res.StatusCode == 404 {
```
khaledkbadr: http module instead of 404

api/node/node_api_MigrateVM.go
```diff
@@ -56,6 +56,13 @@ func (api NodeAPI) MigrateVM(w http.ResponseWriter, r *http.Request) {
...
+		errmsg := fmt.Errorf("node %s does not exist", reqBody.Nodeid)
+		tools.WriteError(w, http.StatusBadRequest, errmsg, "")
+		return
+	}
+
```
khaledkbadr: Add another check for unexpected errors

Issue #1006, opened 2017-09-11
Create Vm with non exist  bridge return 500 instead of 400 


Dinaamagdy:
## Steps 
- Create VM with none existing bridge in it's nic.

## Issue 
it return internal server error instead of  400 bad request bridge not found 

## Software Version 
65b2ed795cc4054e87544ea3b478e3c5aaface31

ahmedelsayed-93:
### software version
master : 851d41fa

PR #1026, opened 2017-09-13
more handling for core0 error to proppagate to api

abdulrahmantkhalifa:
this relates to issue https://github.com/zero-os/0-orchestrator/issues/1006
this depends on PR https://github.com/zero-os/0-core/pull/501

### DIFF
--- a/api/tools/ays.go
+++ b/api/tools/ays.go
@@ -119,25 +119,32 @@ func (aystool AYStool) WaitJobDone(jobid, repoName string) (ays.Job, error) {
 			return job, err
 		}
 	}
+	return aystool.ParseJobError(job.Key, repoName)
+}
 
-	if job.State == "error" {
-		err := AYSError{}
-		if jsonErr := json.Unmarshal([]byte(job.Result), &err); jsonErr != nil {
-			return job, jsonErr
-		}
+// I dont pass the job struct becasue ays does not pass the result even when in error unless a getjob api call is made on it.
+// this will be fixed in ays 9.1.3 ?
+func (aystool AYStool) ParseJobError(jobKey string, repoName string) (ays.Job, error) {
 
-		err.err = fmt.Errorf(err.Message)
-		errResp := http.Response{
-			StatusCode: err.Code,
-		}
-		httperror := HTTPError{
-			Resp: &errResp,
-			err:  err.err,
-		}
-		fmt.Println(httperror)
-		return job, httperror
+	job, _, ayserr := aystool.Ays.GetJob(jobKey, repoName, nil, nil)
+	if ayserr != nil {
+		return job, ayserr
 	}
-	return job, nil
+
+	err := AYSError{}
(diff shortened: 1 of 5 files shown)


api/tools/ays.go
```diff
@@ -119,24 +119,33 @@ func (aystool AYStool) WaitJobDone(jobid, repoName string) (ays.Job, error) {
...
+
+	err.err = fmt.Errorf(err.Message)
+	errResp := http.Response{
+		StatusCode: err.Code,
+	}
+	httperror := HTTPError{
+		Resp: &errResp,
+		err:  err.err,
+	}
+	fmt.Println(httperror)
```
khaledkbadr: remove or use log if logging was the purpose

api/tools/ays.go
```diff
@@ -227,16 +236,16 @@ func (aystool AYStool) archiveBlueprint(blueprintName string, repoName string) e
 func (aystool AYStool) getRun(runid, repoName string) (*ays.AYSRun, error) {
 	run, resp, err := aystool.Ays.GetRun(runid, repoName, nil, nil)
 	if err != nil {
-		return nil, NewHTTPError(resp, err.Error())
+		return &run, NewHTTPError(resp, err.Error())
```
khaledkbadr: if error occurred, the run is useless here no?

api/tools/runs.go
```diff
@@ -30,23 +32,48 @@ func WaitOnRun(api API, w http.ResponseWriter, r *http.Request, runid string) (R
...
+					break
+				}
+			}
 		}
-		return Run{}, err
 	}
+	if jobErr != nil {
+		httpErr, ok := jobErr.(HTTPError)
+		if ok {
+			fmt.Println(httpErr.Resp.StatusCode)
```
khaledkbadr: use log if logging was the purpose

api/tools/runs.go
```diff
@@ -30,23 +32,48 @@ func WaitOnRun(api API, w http.ResponseWriter, r *http.Request, runid string) (R
...
+		errmsg := fmt.Sprintf("error waiting on job %s", job.Key)
+		WriteError(w, http.StatusInternalServerError, err, errmsg)
+		return Run{Runid: run.Key, State: EnumRunState(run.State)}, jobErr
+	}
+
 	if EnumRunState(runstatus.State) != EnumRunStateok {
 		err = fmt.Errorf("Internal Server Error")
 		WriteError(w, http.StatusInternalServerError, err, "")
-		return Run{}, err
+		return Run{Runid: run.Key, State: EnumRunState(run.State)}, jobErr
```
khaledkbadr: not used no?

api/tools/ays.go
abdulrahmantkhalifa: so i can get the errored jobs

api/tools/runs.go
abdulrahmantkhalifa: used cannot pass nil, and i might as well pass the actual run struct instead of an empty one

PR #1039, opened 2017-09-15
Allow API to wait for completion of a run using callback

zaibon:


zaibon:
Closing for now, need to be more tested before merge

Issue #1017, opened 2017-09-13
0-stor configuration needs to validated when the configuration service is created


FastGeert:
Currently the invalid configuration only pops up when a first zero-stor is created.

On top of that, the error message should be improved explaining how the configuration is valid.
```
[Wed13 09:45] - Run.py              :108 :j.atyourservice.server         - ERROR    - error during execution of step 6 in run 56b61a612ef3dd65b262234105a88869
Error of job: storage_cluster!zero-store-cluster (install):
*TRACEBACK*********************************************************************************
Traceback (most recent call last):
  File "/opt/jumpscale9/lib/concurrent/futures/thread.py", line 55, in run
    result = self.fn(*self.args, self.kwargs)
  File "/tmp/actions/storage_cluster/4cbb9221fd4ff204de99461e0537ae6d.py", line 241, in install
    save_config(job)
  File "/tmp/actions/storage_cluster/4cbb9221fd4ff204de99461e0537ae6d.py", line 51, in save_config
    raise RuntimeError("Invalid itsyouonline configuration")
RuntimeError: Invalid itsyouonline configuration


****************************************************************************************
```

PR #1042, opened 2017-09-17
moved validation code into the config service

abdulrahmantkhalifa:
moved the validation code into the config service instead of the storage
cluster creation
relates to issue #1017

### DIFF
--- a/templates/storage_cluster/actions.py
+++ b/templates/storage_cluster/actions.py
@@ -2,6 +2,7 @@ from js9 import j
 
 
 def input(job):
+    from zeroos.orchestrator.configuration import get_configuration
     for arg in ['filesystems', 'arbds']:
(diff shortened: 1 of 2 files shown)


abdulrahmantkhalifa:
should i also validate before installing zero node or is the configuration service enough , if u dont know please dont speculate ask someone who knows

templates/storage_cluster/actions.py
```diff
@@ -269,24 +269,6 @@ def save_config(job):
     zstor_clientid = aysconfig["0-stor-clientid"]
     zstor_clientsecret = aysconfig["0-stor-clientsecret"]
```
khaledkbadr: Can you add a check in the `input` action if the cluster type is of type `object` that give the user a nice error of you can't create object storage cluster without these configs?

Issue #588, opened 2017-07-03
Monitor freediskspace in filesystems


FastGeert:
If only 10% free disks space, set status to warning.
When full set status to error.

ahmedelsayed-93:
### software version
master : 5c7668358

PR #1046, opened 2017-09-17
added monitorint action on filesystem service to check on size

abdulrahmantkhalifa:
relates to issue #588


templates/filesystem/actions.py
```diff
@@ -24,6 +26,7 @@ def install(job):
     except ValueError:
         fs = pool.create(fsname, int(job.service.model.data.quota))
     job.service.model.data.mountpoint = fs.path
+    j.tools.async.wrappers.sync(job.service.executeAction('monitor', context=job.context))
```
khaledkbadr: call monitor(job) directly

templates/filesystem/actions.py
```diff
@@ -38,3 +41,17 @@ def delete(job):
...
+
+
+def monitor(job):
+    service = job.service
+    pool = get_pool(job)
+    for device in pool.fsinfo['devices']:
+        usage_precentage = (device.get('used')/device.get('size'))*100
+        if usage_precentage == 90:
+            service.model.data.status = 'warning'
+        elif usage_precentage > 99:
```
khaledkbadr: @delandtj Is 99 here accurate?

Issue #556, opened 2017-06-30
Import / Export virtual machines


FastGeert:
Goal  is to be able to export a VM from a certain location and import it in a new one.

Requirements:
- export :
  - VM has to be stopped before it can be exported
  - Export is done using [restic](https://github.com/restic/restic)
  - client need to specify a backend where to store the exported VM
    - proposition for this is to provide a REST server that support restic as a service for the client so it's easy for them to setup a storate server for their exported VM. Restic has a plugin for caddy that we could integrate with our IYO plugin to have a complete product without the need for us to implement a lot. see https://github.com/restic/caddy
- import: 
  - client provide URI of the VM he wants to import
  - Restic supports fuse mount, we could use this to expose the exported VM locally in the new location and create and restor the vdisk of the VM from the files in restics.

FastGeert:
After this FR was created, there has been decided that we would use FTP for exporting vdisks. So we will also use FTP(S) for exporting vms.

Dinaamagdy:
Can't export VM and ays return this error : 

```
[Thu12 15:03] - Job.py              :68  :j.core.jobcontroller.job.7deede9a56927eb9b9e57db04f0fff0e - ERROR    - job: vdisk!vdisk3 (export) failed:
Traceback (most recent call last):
  File "/tmp/actions/vdisk/face1038f98d7c6653da594a1d3790ba.py", line 107, in export
    container.waitOnJob(container_job)
  File "/usr/local/lib/python3.5/dist-packages/zeroos/orchestrator/sal/Container.py", line 257, in waitOnJob
    resp.stream(callback)
  File "/usr/local/lib/python3.5/dist-packages/zeroos/core0/client/client.py", line 234, in stream
    callback(meta >> 16, line, meta & 0xff)
  File "/usr/local/lib/python3.5/dist-packages/zeroos/orchestrator/sal/Container.py", line 255, in callback
    raise RuntimeError(erroMessage)
RuntimeError: Flags:   -b, --blocksize int                 the size of the exported (deduped) blocks (default 131072)   -c, --compression CompressionType   the compression type to use, options { lz4, xz } (default lz4)       --conf
ig SourceConfig           config resource: dialstrings (etcd cluster) or path (yaml file) (default config.yml)   -f, --force                         when given, overwrite a deduped map if it can't be loaded   -h, --help            
[...]

PR #1047, opened 2017-09-18
Implement import/export of VMs

khaledkbadr:
related to #556


docs/import_export_vms/README.md
```diff
@@ -0,0 +1,44 @@
+# Export VM
+
+### Requirements:
+* To export a vm, FTP server should be already set up
+
+### Flow
+* First FTP server url should be passed to the Orchestrator API like it's described in [api docs](https://rawgit.com/zero-os/0-orchestrator/master/raml/api.html)
```
abdulrahmantkhalifa: maybe the build script for the ftp can be refrenced in the docs

raml/types.raml
```diff
@@ -28,6 +28,52 @@ CPUInfo:
...
+      description: |
+          FTP server url to export the vm from, and it should be in one of the following formats
+          - localhost:22
+          - ftp://1.2.3.4:200
+          - ftp://user@127.0.0.1:200
+          - ftp://user:pass@12.30.120.200:3000
+          - ftp://user:pass@12.30.120.200:3000/root/dir
+
+
+ExportResUrl:
```
abdulrahmantkhalifa: ExportResUrl is a duplicate from the ExportUrl type use only one

raml/api.raml
```diff
@@ -1259,6 +1259,36 @@ types:
...
                 description: Virtual machine migration request passed successfully
+        /export:
+          post:
+            (AYSBased):
+            securedBy: [itsyouonline: { scopes: [ "user:memberof:0-orchestrator" ] }]
+            displayName: ExportVM
+            description: Export the virtual machine to ftp server
+            body:
+              application/json:
+                type: ExportUrl
```
abdulrahmantkhalifa: this is duplicate type use the same type

templates/vdisk/actions.py
```diff
@@ -286,6 +286,101 @@ def rollback(job):
...
+    service = job.service
+
+    save_config(job)
+
+    url = service.model.data.backupUrl.split("#")[0]
+    parsed_url = urlparse(url)
+    metadata = os.path.basename(parsed_url.path)
+    url = parsed_url.geturl().split(metadata)[0]
+
+    cryptoKey = service.model.data.backupUrl.split("#")[1]
```
abdulrahmantkhalifa: cant we use a library for the URL parsing?

templates/vm/schema.capnp
```diff
@@ -1,4 +1,4 @@
-@0xc32684ebe1b5ced0;
+@0xad54e032f439be18;
```
abdulrahmantkhalifa: why ?

api/node/node_api_ExportVM.go
```diff
@@ -0,0 +1,89 @@
...
+
+	obj := make(map[string]interface{})
+	obj[fmt.Sprintf("vm__%s", vmID)] = bp
+
+	_, err = aysClient.ExecuteBlueprint(api.AysRepo, "vm", vmID, "export", obj)
+	errmsg := fmt.Sprintf("error executing blueprint for vm %s export", vmID)
+	if !tools.HandleExecuteBlueprintResponse(err, w, errmsg) {
+		return
+	}
+
```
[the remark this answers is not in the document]
khaledkbadr: It does the job in processchange, I noticed it finishes the job without wait on run

templates/vdisk/actions.py
khaledkbadr: That's extra info I need for the metadatafile to be created, since I can't know the metadata file name from the actions since it doesn't return values

templates/vm/schema.capnp
khaledkbadr: I changed the schema here, it's good practice to change the ID when you change the schema. Without the change ays must be restarted to pick them up

api/node/node_api_ExportVM.go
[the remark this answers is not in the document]
abdulrahmantkhalifa: if it doesn't need to create a run, u can use aysClient.UpdateBlueprint like the one used in https://github.com/zero-os/0-orchestrator/blob/vm-export/api/node/node_api_UpdateContainer.go#L55

Issue #1055, opened 2017-09-18
Introduce vdisk store concept


zaibon:
Currently the creation of a vdisk though the API requires we specify 3 storage clusters. 
- primary storage cluster for blocks
- slave storage cluster, backup of primary
- object storage cluster for tlog data.

This is becoming cumbersome for user. Proposal is to introduce a new object that aggregate the 3 storage cluster into one logical entity, we would call it vdisk store.
This would link a vdisk to the 3 storage cluster in a easier way.

Also regarding these issue on the 0-disk projects (https://github.com/zero-os/0-Disk/issues/463,  https://github.com/zero-os/0-Disk/issues/462, https://github.com/zero-os/0-Disk/issues/460)
Since we are going to download the templates during installation now, having this vdisk store make sense, cause then we can list the templates available in this vdisk store.

FastGeert:
Should we also list vdisks as children of vdiskstores?

zaibon:
> Should we also list vdisks as children of vdiskstores?

I just checked and currently seems vdisk has no parent define, So I would say yes. It make senses that the vdisk services are located under the vdisk store they belongs.

PR #1076, opened 2017-09-24
Vdiskstorage

abdulrahmantkhalifa:
relates to issue https://github.com/zero-os/0-orchestrator/issues/1055


templates/vdiskstorage/actions.py
```diff
@@ -0,0 +1,14 @@
+from js9 import j
+
+def input(job):
```
khaledkbadr: Add a check if block servers are larger than object servers
```if block_st.model.data.nrServer < object_st.model.data.nrServer:
            raise RuntimeError("blockStoragecluster's number of servers should be equal or larger than them in objectStoragecluster")
```

api/vdiskstorage/CreateNewVdiskStorage.go
```diff
@@ -0,0 +1,71 @@
...
+	if err := reqBody.Validate(); err != nil {
+		tools.WriteError(w, http.StatusBadRequest, err, "")
+		return
+	}
+
+	// validate vdiskstorage name
+	if exists, err := aysClient.ServiceExists("vdiskstorage", reqBody.ID, api.AysRepo); err != nil {
+		tools.WriteError(w, http.StatusInternalServerError, err, "Error checking vdiskstorage service exists")
+		return
+	} else if exists {
```
khaledkbadr: You don't need the else

api/vdiskstorage/CreateNewVdiskStorage.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	if exists {
+		err = fmt.Errorf("vdiskstorage with name %s does  exists", reqBody.ID)
+		tools.WriteError(w, http.StatusBadRequest, err, "")
+		return
+	}
+
+	// validate block cluster name
+	_, response, err := aysClient.Ays.GetServiceByName(reqBody.BlockCluster, "storage_cluster", api.AysRepo, nil, nil)
```
zaibon: What's the idea here ? Just check if the service exists ?
Then why not use `ServiceExists` instead ? I't s a cheaper call for AYS to serve.

Also these 3 calls are not related so you can do them concurrently.

Another way to do it would be to do a list service and filter on the `storage_cluster` role. Then you receive a list of service and you can validate that the service name received in the request actually exists.

api/vdiskstorage/vdiskstorage_if.go
```diff
@@ -0,0 +1,29 @@
...
+	// ListVdiskStorages is the handler for GET /vdiskstorage
+	// List vdisks storages
+	ListVdiskStorages(http.ResponseWriter, *http.Request)
+	// CreateNewVdiskStorage is the handler for POST /vdiskstorage
+	// Create a new vdisk storage, can be a copy from an existing vdisk
+	CreateNewVdiskStorage(http.ResponseWriter, *http.Request)
+}
+
+// VdiskstorageInterfaceRoutes is routing for /vdiskstorage root endpoint
+func VdiskstorageInterfaceRoutes(r *mux.Router, i VdiskstorageInterface, org string) {
```
zaibon: Why the `org` argument ?

raml/types.raml
```diff
@@ -1105,6 +1105,22 @@ VdiskCreate:
...
+  properties:
+    id: string
+    blockCluster:
+      type: string
+      description: block cluster to assign to vdisk store
+      required: True
+    slaveCluster:
+      type: string
+      description: slave cluster to assign to vdisk store
+      required: True
```
zaibon: I don't think this is required, we could have a vdisk storage with only a blockCluster and no slave or object cluster

pyclient/zeroos/orchestrator/client/VdiskStorage.py
```diff
@@ -0,0 +1,87 @@
...
+from . import client_support
+
+
+class VdiskStorage(object):
+    """
+    auto-generated. don't touch.
+    """
+
+    @staticmethod
+    def create(blockCluster, id, objectCluster=None, slaveCluster=None):
```
muhamadazmy: It's weird that u keep putting the ID as a second argument. Is there a reason for that?

api/vdiskstorage/CreateNewVdiskStorage.go
```diff
@@ -0,0 +1,96 @@
...
+		if err != nil {
+			tools.WriteError(w, http.StatusInternalServerError, err, "Error checking storage_cluster service exists")
+			return
+		}
+		if !exists {
+			err = fmt.Errorf("storage_cluster with name %s does not exists", reqBody.ObjectCluster)
+			tools.WriteError(w, http.StatusBadRequest, err, "")
+			return
+		}
+		if reqBody.SlaveCluster != "" {
```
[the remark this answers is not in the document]
abdulrahmantkhalifa: @khaledkbadr  says for slave to exist there must be an object cluster and for object to exist there must a be a block cluster

pyclient/zeroos/orchestrator/client/VdiskStorage.py
abdulrahmantkhalifa: genertaed

api/vdiskstorage/CreateNewVdiskStorage.go
```diff
@@ -0,0 +1,96 @@
...
+	exists, err := aysClient.ServiceExists("vdiskstorage", reqBody.ID, api.AysRepo)
+
+	if err != nil {
+		tools.WriteError(w, http.StatusInternalServerError, err, "Error checking vdiskstorage service exists")
+		return
+	}
+
+	if exists {
+		err = fmt.Errorf("vdiskstorage with name %s already exists", reqBody.ID)
+		tools.WriteError(w, http.StatusBadRequest, err, "")
```
zaibon: This should return a `http.StatusConflict` and not a bad requests.

api/vdiskstorage/CreateNewVdiskStorage.go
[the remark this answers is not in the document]
zaibon: We have to add a check that if SlaveCluster is specified but object ObjectCluster is emtpy, then it's a bad requets.

templates/tlogserver/actions.py
```diff
@@ -48,11 +48,12 @@ def save_config(job, vdisks=None):
     etcd.put(key="%s:cluster:conf:tlog" % service.name, value=yamlconfig)
 
     for vdisk in vdisks:
+        vdiskstore = vdisk.producers['vdiskstorage'][0]
```
zaibon: this is dangerous, what if `vdisk.producers['vdiskstorage']` is emtpy ?

templates/tlogserver/actions.py
```diff
@@ -80,13 +81,14 @@ def install(job):
 
     backup = False
     for vdiskservice in vdisks:
-        objectcluster = vdiskservice.model.data.objectStoragecluster
+        vdiskstore = vdiskservice.producers['vdiskstorage'][0]
```
zaibon: Idem, need to prevent panic in case list is emtpy

templates/vdisk/actions.py
```diff
@@ -1,18 +1,6 @@
 from js9 import j
 
 
-def input(job):
```
zaibon: Why removing all the input checks ? we still can validate that we have a vdiskStorage consumed etc...

templates/vdisk/actions.py
```diff
@@ -32,8 +20,10 @@ def install(job):
         template = urlparse(service.model.data.templateVdisk)
         targetconfig = get_cluster_config(job)
         target_node = random.choice(targetconfig['nodes'])
-        blockStoragecluster = service.model.data.blockStoragecluster
-        objectStoragecluster = service.model.data.objectStoragecluster
+        vdiskstore = service.producers['vdiskstorage'][0]
```
zaibon: again, check if list is not empty

templates/vdisk/actions.py
```diff
@@ -93,8 +83,10 @@ def save_config(job):
     import yaml
     from zeroos.orchestrator.sal.ETCD import EtcdCluster
     from zeroos.orchestrator.configuration import get_jwt_token
+    service = job.service
 
     job.context['token'] = get_jwt_token(job.service.aysrepo)
+    vdiskstore = service.producers['vdiskstorage'][0]
```
zaibon: empty list

templates/vdisk/actions.py
abdulrahmantkhalifa: it already does, if parent

PR #1104, opened 2017-09-28
Replaced clientid & secret parameters with jwt

FastGeert:
Also improved how jwt's are refreshed into 1 instance, instead of 1 per api section

### DIFF
--- a/api/tools/jwt.go
+++ b/api/tools/jwt.go
@@ -78,7 +78,7 @@ func (jp *JWTProvider) refreshToken() (string, error) {
 	return jp.jwt, nil
 }
 
-func (jp *JWTProvider) GetToken() (string, error) {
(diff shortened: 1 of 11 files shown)


api/tools/jwt.go
```diff
@@ -6,79 +6,82 @@ import (
...
-	claim, err := GetJWTClaim(token)
+	err := jp.doRefreshToken()
 	if err != nil {
 		return "", err
 	}
-	exp := int64(claim["exp"].(float64))
+	return jp.jwt, nil
+}
+
+func (jp *JWTProvider) GetToken() (string, error) {
```
LeeSmet: Since this returns a JWT and not an access token, "GetJWT" would be more descriptive and clear here

PR #1118, opened 2017-09-29
adjusted th sal and actions dhcp to avoid watchdog

abdulrahmantkhalifa:
- added a stop method in the dhcp sal
 - called the stop method in the actions to seperate concerns and
   inssure the dhcp is down before saving status as halted
 - after applying config added status running


pyclient/zeroos/orchestrator/sal/gateway/dhcp.py
```diff
@@ -43,3 +35,14 @@ def is_running(self):
...
+    def stop(self):
+        for process in self.container.client.process.list():
+            if 'dnsmasq' in process['cmdline']:
+                self.container.client.process.kill(process['pid'], signal.SIGTERM)
+                start = time.time()
+                while start + 10 > time.time():
+                    if not self.container.client.process.list(process['pid']):
+                        break
+                    time.sleep(0.2)
+                break
```
rkhamis: Shouldn't we raise an error if we fail to stop? Or do a hard kill or something? Wouldn't this just fail silently?

FastGeert: Replace SIGTERM with SIGKILL and remove the entire loop to see if it's still there

PR #1121, opened 2017-10-01
Import images

zaibon:


:
```diff
@@ -0,0 +1,206 @@
+from js9 import j
```
zaibon: What is this file ?

api/vdiskstorage/vdiskstorage_api_DeleteImage.go
```diff
@@ -0,0 +1,64 @@
...
+	vars := mux.Vars(r)
+	imageID := vars["imageid"]
+
+	exists, err := aysClient.ServiceExists("vdisk_image", imageID, api.AysRepo)
+	if err != nil {
+		tools.WriteError(w, http.StatusInternalServerError, err, "Failed to check for the image")
+		return
+	}
+
+	if !exists {
```
zaibon: Since it's a delete action, we could return a success if the image doesn't exist instead of a not found

api/vdiskstorage/vdiskstorage_api_ListImages.go
```diff
@@ -0,0 +1,70 @@
...
+
+	ays "github.com/zero-os/0-orchestrator/api/ays-client"
+	"github.com/zero-os/0-orchestrator/api/tools"
+)
+
+// ListImages is the handler for GET /vdiskstorage/{vdiskstorageid}/images
+// List all vdisk images installed in this VdiskStroage
+func (api *VdiskstorageAPI) ListImages(w http.ResponseWriter, r *http.Request) {
+	aysClient := tools.GetAysConnection(r, api)
+	// vdiskstorageID := mux.Vars(r)["vdiskstorageid"]
```
zaibon: remove commented line

templates/vdisk_image/actions.py
```diff
@@ -0,0 +1,215 @@
...
+                                                               instance=vdiskstore.model.data.objectCluster)
+            cmd += ' --data-shards {} --parity-shards {}'.format(storageclusterservice.model.data.dataShards,
+                                                                 storageclusterservice.model.data.parityShards)
+
+        job.logger.info("import image {} from {} as {}".format(snapshotID, url, service.name))
+        job.logger.info(cmd)
+
+        container_job = container.client.system(cmd, id="vdisk.import.%s" % service.name)
+
+        try:
```
zaibon: Should this be in a try except block?

templates/vm/actions.py
```diff
@@ -256,9 +256,10 @@ def install(job):
             service.saveAll()
             raise j.exceptions.Input(str(e))
 
+
         # wait for max 60 seconds for vm to be running
         start = time.time()
-        while start + 60 > time.time():
+        while start + 600 > time.time():
```
zaibon: put back to 60 ?

:
FastGeert: Must be a mistake. Removed it.

api/vdiskstorage/vdiskstorage_api_DeleteImage.go
FastGeert: Agree.

templates/vdisk_image/actions.py
FastGeert: I don't like these sort of try except blocks, as it will hide the real problem

templates/vm/actions.py
FastGeert: yes

PR #1123, opened 2017-10-01
Return error to the user if authentication failed

khaledkbadr:


api/tools/connection.go
```diff
@@ -139,16 +136,16 @@ func ConnectionMiddleware(opt ...ConnectionOptions) func(h http.Handler) http.Ha
...
-func GetAysConnection(r *http.Request, api API) AYStool {
+func GetAysConnection(r *http.Request, api API) (AYStool, error) {
 	aysAPI := api.AysAPIClient()
 
 	token, err := api.GetJWT()
 	if err != nil {
-		log.Error(err.Error())
-	} else {
-		aysAPI.AuthHeader = fmt.Sprintf("Bearer %s", token)
+		return GetAYSClient(aysAPI), err
```
FastGeert: Should return nil, err

api/tools/jwt.go
```diff
@@ -88,8 +87,12 @@ func (jp *JWTProvider) refreshToken() (string, error) {
 }
 
 func (jp *JWTProvider) GetJWT() (string, error) {
+	if jp.development {
+		return jp.jwt, nil
+	}
+
 	now := time.Now().Unix()
-	if rem := jp.expires - now; rem < 300 {
+	if rem := jp.expires - now; rem < 3600 {
```
FastGeert: ??? JWT are refreshed for 3600 seconds, which always result in less than 3600 seconds remaining. The original code is correct.

api/tools/jwt.go
```diff
@@ -88,8 +87,12 @@ func (jp *JWTProvider) refreshToken() (string, error) {
 }
 
 func (jp *JWTProvider) GetJWT() (string, error) {
+	if jp.development {
+		return jp.jwt, nil
```
FastGeert: Shouldn't we return "", nil ??? there is no jwt, so

rkhamis: It's already going to be empty, since we get the DevelmentJWTProvider

api/tools/connection.go
khaledkbadr: The method return `AYStool` object, so I thought to return the default one. If you have an idea how to go around that to return nil Im all for it

FastGeert: The nil is for the error case

api/tools/jwt.go
FastGeert: indeed, but it does not make sense to return jp.jwt. People that will read the code later on, will then clearly see that it always returns "" in development mode.

PR #1162, opened 2017-10-04
Tlog block size

zaibon:


templates/node.zero-os/actions.py
```diff
@@ -356,6 +356,10 @@ def watchdog(job):
     }
 
     async def callback(jobid, level, message, flag):
+        if "nbd" in jobid:
```
muhamadazmy: Do u still need this, looks like a debug code to me

templates/vdisk_image/actions.py
```diff
@@ -29,6 +29,7 @@ def install(job):
         cmd = "/bin/zeroctl import vdisk {vdiskid} {snapshotID} -j 20\
                --config {dialstrings} \
                --blocksize {blocksize} \
```
muhamadazmy: Shouldn't the blocksize be removed ? or i miss understood Glen?

muhamadazmy: Shouldn't blocksize be removed?

templates/node.zero-os/actions.py
zaibon: @FastGeert  ?

templates/vdisk_image/actions.py
zaibon: I think it should indeed since it doesn't exist on master branch of 0-disk anymore.
@FastGeert ?

PR #1176, opened 2017-10-09
Version bug fix

zaibon:


templates/node.zero-os/actions.py
```diff
@@ -12,14 +12,17 @@ def get_statsdb(service):
...
 def get_version(job):
     from zeroos.orchestrator.sal.Node import Node
     from zeroos.orchestrator.configuration import get_jwt_token
     service = job.service
     node = Node.from_ays(service, get_jwt_token(job.service.aysrepo))
-    service.model.data.version = node.client.ping().split('Version:')[1]
+
+    pong = node.client.ping()
+    version = pong.split('Version:')[1] if pong else ''
+    service.model.data.version = version if version else service.model.data.version
```
zaibon: I don't get this line

templates/node.zero-os/actions.py
```diff
@@ -133,7 +137,7 @@ def monitor(job):
         try:
             node = Node.from_ays(service, token, timeout=5)
             node.client.testConnectionAttempts = 0
-            state = node.client.ping()
+            state = get_version(job)
```
zaibon: not sure why get_version is assigned to a state variable.

abdulrahmantkhalifa: it is the same as the pong it just returns a string , as far as i can tell it is just used as bool

zaibon: This is not correct then. This should only do a ping. 
the `get_version` can return a not empty string even if the ping failed

PR #1182, opened 2017-10-09
Disaster recovery

zaibon:
- remove async wrapper
- add some disaster recovery code
- lots of improvement regarding live migration of vms


templates/cloudinit/actions.py
```diff
@@ -55,8 +55,4 @@ def watchdog_handler(job):
     loop = j.atyourservice.server.loop
     gateway = job.service.parent.consumers['gateway'][0]
     if gateway.model.data.status == 'running':
-        asyncio.ensure_future(job.service.executeAction('start', context=job.context), loop=loop)
-
-
-def monitor(job):
-    pass
+        asyncio.ensure_future(job.service.ayncExecuteAction('start', context=job.context), loop=loop)
```
khaledkbadr: typo, shouldn't it be `asyncExecuteAction`

templates/storage_cluster/actions.py
```diff
@@ -409,21 +409,42 @@ def delete(job):
...
+        container.delete()
+
+    for zerostor in zerostors:
+        tcps = zerostor.producers.get('tcp', [])
+        for tcp in tcps:
+            tcp.executeAction('drop', context=job.context)
+            tcp.delete()
+
+        container = zerostor.parent
+        container.executeAction('stop', context=job.context)
```
rkhamis: Stop will happen anyway, you won't need to schedule it. Because you don't overwrite `init_actions_`
Default `init_actions_` already has `delete` dependent on `uninstall` which is dependent on `stop`

zaibon: That's not true, what you're talking about is valild when creating a run. Here we just execute an action alone.

templates/node.zero-os/actions.py
```diff
@@ -307,6 +311,7 @@ def watchdog(job):
     watched_roles = {
         'nbdserver': {
             # 'message': (re.compile('^storageengine-failure.*$')),  # TODO: Not implemented yet in 0-disk yet
+            'message': (re.compile('.*'),),
```
khaledkbadr: Filter on desired messages or part of messages only. This way the watchdog_hanlder will be called on every message which make ays slow

PR #1192, opened 2017-10-10
Delete vdisk etcd

abdulrahmantkhalifa:


templates/vdisk/actions.py
```diff
@@ -138,6 +139,33 @@ def save_config(job):
...
+    etcd_cluster = service.aysrepo.servicesFind(role='etcd_cluster')[0]
+    etcd = EtcdCluster.from_ays(etcd_cluster, job.context['token'])
+
+    # delete base config
+    etcd.delete(key="%s:vdisk:conf:static" % service.name)
+
+    # delete tlog config from etcd
+    vdiskType = service.model.data.type
+    objectStoragecluster = '' if vdiskType == 'tmp'or vdiskType == 'cache' else vdiskstore.model.data.objectCluster
+    if objectStoragecluster:
```
khaledkbadr: why not check `if vdiskType == 'tmp'or vdiskType == 'cache'` and drop the vdiskstore variable. Or even check on `if vdiskstore.model.data.objectCluster`?

Issue #1016, opened 2017-09-13
ETCD cluster installation should prefer to use NVME devices if available


FastGeert:
The etcd key value store should have the fastest storage available in the system

abdulrahmantkhalifa:
is the PR what u had in mind @FastGeert

abdulrahmantkhalifa:
#1203

ahmedelsayed-93:
### software version
master : 851d41fa

PR #1203, opened 2017-10-12
etcd now will prioritize faster disks for its sp

abdulrahmantkhalifa:
it will search for NVME then SSD then will default to fscache whatever
that may be. do not handle using ready sp unless they are old etcd sp
created from before.

relates to issue #1016

### DIFF
--- a/pyclient/zeroos/orchestrator/utils.py
+++ b/pyclient/zeroos/orchestrator/utils.py
@@ -27,7 +27,7 @@ def find_disks(disk_type, nodes, partitionName):
         for disk in node.disks.list():
             # skip disks of wrong type
             if disk.type.name != disk_type:
-                    continue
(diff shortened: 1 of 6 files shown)


pyclient/zeroos/orchestrator/utils.py
```diff
@@ -7,3 +7,36 @@ def Write_Status_code_Error(job, exception):
...
+        for partition in disk.partitions:
+            for filesystem in partition.filesystems:
+                if filesystem['label'].startswith(partitionName):
+                    return True
+
+    for node in nodes:
+        for disk in node.disks.list():
+            # skip disks of wrong type
+            if disk.type.name != disk_type:
+                    continue
```
FastGeert: IndentationError! :)

PR #1223, opened 2017-10-17
Master gateway migrate

### LINKED ISSUE #1132: Migrate gateway to other node
Need an action on the rest-api that can move a gateway to another node without interrupting it.

Basically it means that for a very brief time the gateway will have an instance on both the source and target node, to limit the amount of network interruption. 

@delandtj is there some optimization that can be done on the network level to make the transition as shot as possible?

muhamadazmy:
Fixes #1132 

Support migrating gateway to a different node

### DIFF
--- a/api/node/node_api_MigrateGW.go
+++ b/api/node/node_api_MigrateGW.go
@@ -73,5 +73,4 @@ func (api *NodeAPI) MigrateGW(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Set("Location", fmt.Sprintf("/nodes/%s/gws/%s", reqBody.Node, gwname))
 	w.WriteHeader(http.StatusOK)
-
 }
--- a/api/node/node_api_MigrateGW.go
+++ b/api/node/node_api_MigrateGW.go
@@ -1,0 +1,80 @@
+package node
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/gorilla/mux"
+
+	"github.com/zero-os/0-orchestrator/api/tools"
+)
+
+// CreateGW is the handler for POST /nodes/{nodeid}/gws
+// Create a new gateway
+func (api *NodeAPI) MigrateGW(w http.ResponseWriter, r *http.Request) {
+	aysClient, err := tools.GetAysConnection(api)
+	if err != nil {
+		tools.WriteError(w, http.StatusUnauthorized, err, "")
+		return
+	}
+
+	var reqBody MigrateGW
+
+	vars := mux.Vars(r)
+	gwname := vars["gwname"]
+
+	// decode request
+	if err := json.NewDecoder(r.Body).Decode(&reqBody); err != nil {
+		tools.WriteError(w, http.StatusBadRequest, err, "Error decoding request body")
+		return
+	}
+
+	// validate request
+	if err := reqBody.Validate(); err != nil {
+		tools.WriteError(w, http.StatusBadRequest, err, "")
+		return
+	}
+
+	exists, err := aysClient.ServiceExists("gateway", gwname, api.AysRepo)
+	if err != nil {
+		errmsg := fmt.Sprintf("error getting gateway service by name %s ", gwname)
+		tools.WriteError(w, http.StatusInternalServerError, err, errmsg)
+		return
+	}
+	if !exists {
+		tools.WriteError(w, http.StatusNotFound, fmt.Errorf("gateway with name %s does not exist", gwname), "")
+		return
+	}
+
+	exists, err = aysClient.ServiceExists("node", reqBody.Node, api.AysRepo)
+	if err != nil {
+		errmsg := fmt.Sprintf("error getting node service by name %s ", reqBody.Node)
+		tools.WriteError(w, http.StatusInternalServerError, err, errmsg)
+		return
+	}
+	if !exists {
+		tools.WriteError(w, http.StatusNotFound, fmt.Errorf("node with name %s does not exist", reqBody.Node), "")
+		return
+	}
+
+	bp := map[string]interface{}{
+		fmt.Sprintf("gateway__%s", gwname): map[string]string{
+			"node": reqBody.Node,
+		},
+	}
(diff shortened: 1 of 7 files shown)


templates/gateway/actions.py
```diff
@@ -218,12 +218,26 @@ def processChange(job):
 
     gatewaydata = service.model.data.to_dict()
     nicchanges = gatewaydata['nics'] != args.get('nics')
-    httproxychanges = gatewaydata['httpproxies'] != args.get('httpproxies')
-    portforwardchanges = gatewaydata['portforwards'] != args.get('portforwards')
+    httproxychanges = gatewaydata.get('httpproxies') != args.get('httpproxies')
```
grimpy: if you do this the rest of the update will think all httproxies have been removed same for line underneed with portforwards

templates/gateway/actions.py
```diff
@@ -218,12 +218,26 @@ def processChange(job):
...
+    portforwardchanges = gatewaydata.get('portforwards') != args.get('portforwards')
+    nodechanged = gatewaydata.get('node') != args.get('node')
 
     container = service.producers.get('container')[0]
     containerobj = Container.from_ays(container, job.context['token'], logger=job.service.logger)
 
+    if nodechanged:
+        node = service.aysrepo.servicesFind(role='node', name=args.get('node'), first=True)
```
zaibon: use `serviceGet` here and handle the case where the services doesn't exists

templates/gateway/actions.py
muhamadazmy: I see, should i move the migrate logic to a separate action ?

api/node/node_api_MigrateGW.go
```diff
@@ -0,0 +1,80 @@
...
+		return
+	}
+
+	bp := map[string]interface{}{
+		fmt.Sprintf("gateway__%s", gwname): map[string]string{
+			"node": reqBody.Node,
+		},
+	}
+	//aysClient.UpdateBlueprint()
+	run, err := aysClient.ExecuteBlueprint(api.AysRepo, "gateway", gwname, "update", bp)
```
zaibon: This is triggerd from the process change so you need to use the `updateBlueprint` method and not wait on a run

templates/gateway/actions.py
```diff
@@ -218,12 +218,26 @@ def processChange(job):
...
+        service.executeAction('stop', context=job.context)
```
zaibon: In the issue it's said to not have downtime so both gateway need to be up at the same time for a short period. 
This is not what happens here right ?

templates/gateway/actions.py
grimpy: thats not possible as processChanges is the only one able to detect changes.
I suggest you put the line of nodechanged on top and move your if underneed and do a return there.
Because when you change the node all the other stuff isnt required as I can see from the PR you are not updating nics or portforwards or httpproxies during move.

templates/gateway/actions.py
muhamadazmy: No, i stop the gateway before i migrate it.

api/node/node_api_MigrateGW.go
muhamadazmy: I was confused since the UpdateBlueprint takes an action as well, also this actually executed the blueprint so i don't get the difference.

templates/gateway/actions.py
muhamadazmy: The serviceFind with a `first=True` will raise an exception (and crash the action) if the node doesn't exist. I believe that a similar behaviour to serviceGet. Will check

templates/gateway/actions.py
muhamadazmy: True
Okay, will do it this way.

api/node/node_api_MigrateGW.go
zaibon: Well I don't really get why there is so many method to do more or less the same either.
But for the processChange method the preferred method is `UpdateBlueprint` any other method will bug on trying to get job state from a run that doesn't exists, cause you don't need to create  a run here.

templates/gateway/actions.py
```diff
@@ -217,6 +217,21 @@ def processChange(job):
         return
 
     gatewaydata = service.model.data.to_dict()
+    nodechanged = gatewaydata.get('node') != args.get('node')
```
zaibon: We are still stopping the gateway during migration, which is not what is asked in #1132 .

Also, this processChange is getting very long and hard to follow.
Why can't we just have in the processChange the code that detect what needs to be done, then it calls an action that actually do the required change. so we have small and readable code.

templates/gateway/actions.py
```diff
@@ -204,6 +204,19 @@ def get_zerotier_nic(zerotierid, containerobj):
         raise j.exceptions.RuntimeError("Failed to get zerotier network device")
 
 
+def migrate(job, service, container, containerobj, dest):
```
zaibon: This signature doesn't make sense to me.
it could simply be `def migrate(job, dest):` all the rest can be deduce from the job
I know it would imply to create a new container sal object. but you only already have it because the `processChange` method is still messy and already creates it

muhamadazmy: Yeah, i didn't want to recreate almost everything, i thought since i have them i can just pass them around, they are references anyway.

zaibon: yes but then it makes the migrate method really strange. I prefer to have something clean and logic then just trying to optimize small stuff like this

muhamadazmy: updated

templates/gateway/actions.py
```diff
@@ -204,6 +204,24 @@ def get_zerotier_nic(zerotierid, containerobj):
         raise j.exceptions.RuntimeError("Failed to get zerotier network device")
 
 
+def migrate(job, dest):
+    from zeroos.orchestrator.sal.Container import Container
+    service = job.service
+    container = service.producers.get('container')[0]
```
zaibon: should check if `service.producers.get('container')` actually return a list that has a item inside.
better safe then sorry.

Issue #1246, opened 2017-10-22
deleting the last device in storage pool returns 204 instead of Bad Request


ahmedelsayed-93:
### API
```DELETE  /nodes/{nodeid}/storagepools/{storagepoolname}/devices/{deviceuuid}```

### Scenario
- create storage cluster with metadataProfile and dataProfile equal to ``single``.
- attach to it only one device.
- try to delete this device.

### expected result
bad request 

#### Api response
```204```

#### AYS error
```
Traceback (most recent call last):
  File "/usr/lib/python3.5/concurrent/futures/thread.py", line 55, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/tmp/actions/storagepool/f3cbf63a3aa9a3cc36e992f84eb303a9.py", line 98, in processChange
    updateDevices(service, pool, devices)
  File "/tmp/actions/storagepool/f3cbf63a3aa9a3cc36e992f84eb303a9.py", line 153, in updateDevices
    pool.device_remove(*removeddevices)
  File "/usr/local/lib/python3.5/dist-packages/zeroos/orchestrator/sal/StoragePool.py", line 147, in device_remove
    self._client.btrfs.device_remove(self._get_mountpoint(), *devices)
  File "/usr/local/lib/python3.5/dist-packages/zeroos/core0/client/client.py", line 1843, in device_remove
    self._client.sync('btrfs.device_remove', args)
  File "/usr/local/lib/python3.5/dist-packages/zeroos/core0/client/client.py", line 760, in sync
    raise ResultError(msg='%s' % result.data, code=result.code)
zeroos.core0.client.client.ResultError: "(ERROR): [ ERROR: error removing device '/dev/sda1': unable to remove the only writeable device\n]"

```

ahmedelsayed-93:
### software version
master : 851d41fa

PR #1271, opened 2017-10-24
Adds return bad request if someone is trying to delete the last devic…

ashraffouda:
Adds return bad request if someone is trying to delete the last device in the pool


ashraffouda:
https://github.com/zero-os/0-orchestrator/issues/1246

Issue #1372, opened 2017-12-18
Installer fails to retrieve the correct zerotier ip address


ahussein:
While trying to install 0-orch using the installer file. the following errors occurred

```
orchestratorapiserver --bind 'fc7b:8c7b:f461:3fb3:e5cb:0000:0000:0001:8080' --ays-url http://127.0.0.1:5000 --ays-repo orchestrator-server --org 'abdel_org' --jwt ''eyJhbGciOiJFUzM4NCIsInR5cCI6IkpXVCJ9.eyJhenAiOiI4S1FsS0JRQWVxRFFvZWZWNTVLTHdDZVI1ZGRRIiwiZXhwIjoxNTEzNjEzNDU2LCJpc3MiOiJpdHN5b3VvbmxpbmUiLCJyZWZyZXNoX3Rva2VuIjoiVkJJOEFJRFJUdTN0UXFxX29Md2diSkRqNmt5XyIsInNjb3BlIjpbIiJdLCJ1c2VybmFtZSI6ImFiZGVscmFobWFuX2h1c3NlaW5fMSJ9.ZlNmhEFJOjkqgQOrbcoI-_7MtFGmexVbiquDPIVriA3asD8xfSqsadXMqnwmvDdirrL7l3PFgizCSc2uc3NoN0Kv-Hb3fuF4sUiLTyDastdp53s1tSegi7xSdF3prUVt'' 
[...]

PR #1373, opened 2017-12-18
fix installer and update docs

ahussein:
Fixes: https://github.com/zero-os/0-orchestrator/issues/1372


ahussein:
@rkhamis can you check this again. I did a little change more.

PR #1374, opened 2018-01-03
Hardened bootstrap service

FastGeert:


templates/bootstrap.zero-os/actions.py
```diff
@@ -21,7 +21,19 @@ def bootstrap(job):
...
+        if member['nodeId'] in service.model.data.authorizedZerotierMembers:
+            if not member['config']['authorized']:
+                job.logger.info("Authorizing %s" % member['nodeId'])
+                member['config']['authorized'] = True
+                zerotier.network.updateMember(member, member['nodeId'], netid)
+        else:
+            to_process.append(member)
+
+    # In a second pass process all the others
+    for member in to_process:
```
ahussein: I dont see where the service.model.data.authorizedZerotierMembers is being updated.

zaibon: Christophe de Carvalho, [03.01.18 13:53]
where do you update service.model.data.authorizedZerotierMembers?

Geert, [03.01.18 13:54]
thats configuration of the bootstrap service

Christophe de Carvalho, [03.01.18 13:54]
?

Christophe de Carvalho, [03.01.18 13:54]
so you add this manually ?

Geert, [03.01.18 13:54]
i manually added the orchestrator node itself

Christophe de Carvalho, [03.01.18 13:54]
it's not filled in when authorizing node ?

Geert, [03.01.18 13:54]
no

Geert, [03.01.18 13:55]
it was not my intention to add the nodes here

Geert, [03.01.18 13:55]
but to add machines that are not zero-os

Geert, [03.01.18 13:55]
that should not get deauthorized

ahussein: okay, thanks @zaibon. I think the docs should be updated accordingly then.

FastGeert: It's a configuration thing. I manually added the orchestrator machine itself here.

